PR #17, opened 2025-10-17
@feat/engine base

ThomasRuijters:
This PR adds:

1. A readme for how to build the engine
2. A CMake user presets with default and examples file
3. A CMakeLists.txt file that builds ENET, Box2D, Freetype, SDL3, SDL3 Image ext, SDL3 ttf ext, ImGui.
4, The required file and folder structure for the engine.**src** for cpp files, **include** for public headers, and lib for private headers and dependencies. The **lib** folder will mainly be used for dependencies we ourselves want to include into the engine, not so much for stuff we import or use under the hood like SDL. The resources folder is for possible resources that we may want to include into the engine by default like default textures or sprite-sheets.


CMakeLists.txt
```diff
@@ -0,0 +1,175 @@
...
+            "${FREETYPE_DLL}"
+            $<TARGET_FILE_DIR:${PROJECT_NAME}>
+    )
+
+    # Copy resources
+    add_custom_command(TARGET ${PROJECT_NAME} POST_BUILD
+            COMMAND ${CMAKE_COMMAND} -E copy_directory
+            ${CMAKE_CURRENT_SOURCE_DIR}/resources
+            $<TARGET_FILE_DIR:${PROJECT_NAME}>/resources
+    )
```
BotSquirrel: My build crashed due to this, i do run a slighty different build using an userpreset but copying a dir that does not exist is likely to cause something?

CMakeLists.txt
```diff
@@ -0,0 +1,175 @@
...
+# =========================================================
+# Sources
+# =========================================================
+file(GLOB_RECURSE SOURCES CONFIGURE_DEPENDS
+        src/*.cpp src/*.h
+        lib/*.cpp lib/*.h
+        include/*.cpp include/*.h)
+
+add_executable(${PROJECT_NAME} ${SOURCES} ${IMGUI_SOURCES})
+
```
BotSquirrel, line 121: The `scene-poc` uses CMake's ability to compile directories as libraries. We could do that here too if everyone agrees? Not much would change and `c++` files would just shift somewhere else.

If we do it then we work from `engine/` and CMake turns it into a library meaning we can just do `#include <engine/core>` instead of `#include "../engine/core"`. 

From what I saw main won't really be used anyway as we make classes that should be used in combination to achieve something, so why not make it our local test file (just don't commit your tests here)?

```CMake
# =========================================================
# Engine library
# =========================================================
file(GLOB_RECURSE ENGINE_SOURCES CONFIGURE_DEPENDS src/*.cpp)

add_library(engine ${ENGINE_SOURCES} ${IMGUI_SOURCES})

target_include_directories(engine
    PUBLIC ${CMAKE_CURRENT_SOURCE_DIR}/include
    PRIVATE ${imgui_SOURCE_DIR} ${imgui_SOURCE_DIR}/backends
)

target_link_libraries(engine
    PRIVATE ${SDL_MAJOR}::${SDL_MAJOR} enet box2d
)

target_link_libraries(${PROJECT_NAME} PRIVATE engine ${SDL_MAJOR}::${SDL_MAJOR} enet box2d)

```

BotSquirrel (commented):
Pulled it locally and was able to build. I did get a crash the first time which i placed a comment on. I think its fine but I would like others to see my comments first before approving as it could be something

CMakeLists.txt
ThomasRuijters: If resources does not exist it will go kaboom. I noticed this as well. Should not be an issue, but we either make it optional or make a default empty folder if none exist.

CMakeLists.txt
ThomasRuijters: I agree. The point is that we do have to work with a resources folder by default. We have not thought of how to work with a user-provided one yet. This could easily be fixed with a config file ey, and last time I checked it is only a problem when you try to compile the engine itself. Users will probably provide defaults themselves, so we do not have to worry about that. So as long as that is not an issue, be my guest and do what you proposed.

ThomasRuijters: What would be the folder structure then? Would it be an engine folder with lib, include, and src in it? What would I expect here? Or would you make a lib for the core, one for common, and one for the public api? Where would the game live? I do not want to turn the entire repository into some kind of mono repo, so what do we do with that? At the end of the day we are going to get two completely separate apps (engine + game)

BotSquirrel: See the [scene-poc](https://github.com/SPC-F/scene-poc), that is how it would look like:
```
|- include
|--|- engine (only .h)
|----|- core
|----|- components
|----|- ...
|- src (reflection of engine but the .cpp)
|--|- core
|--|- components
|- main.cpp
```

`Engine` is pretty much a library already here and we don't need the `main.cpp` as we never initiate something right? I agree that the Game should definitely be build somewhere else, but would it not be nice if you could test something here? Otherwise if you'd want to test something quick you would need to compile this, copy it over, link it with the game and then do your test there.

The file structure is also not set in stone. We can do `include/engine/api` and `include/engine/internal` if we want. This is just what I used as example and found that it worked quite nicely.

CMakeLists.txt
```diff
@@ -0,0 +1,175 @@
+cmake_minimum_required(VERSION 3.25)
```
LarsDevans, line 1: CMake is currently stable on version 4.2.0. Do we want to do something with that?

LarsDevans (approved):

CMakeLists.txt
BotSquirrel: I added this as discussed in our daily standup on (20/10/25) 👌

BotSquirrel (approved):

SjulGr (approved):
I made two commits:
1. Set the library versions of `SDL3`, `SDL3_image` and `SDL3_ttf` to compatible versions. `SDL3_image`'s release tag was set to `main` which caused it to fetch a version incompatible with our `SDL3` version. All three now have the most recent release.
3. Changed the project file name in the `readme.md` from `capycore-engine` to `CapycoreEngine` since the latter is the actual folder name.

PR #19, opened 2025-10-22
feat (engine base classes): Add engine base

### LINKED ISSUE #18: @feat/engine_base
Add:

1. Engine
2. Service container
3. IManager
4. Transform
5. Gameobject
6. Component

ThomasRuijters:
Added:
1. Add ServiceContainer implementation;
2. Add Engine base implementation
3. Add IEngineService implementation
4. Add Point implementation
5. Add Transform implementation
6. Add Vector3 implementation
7. Add Component implementation
8. Add Scene scaffold. Does nothing for now it really is just an empty class
9. Add GameObject base implementation

Changed:
1. A lot of classes implement weak_ptr. Now, that can work from time to time, but we often use it in semantically incorrect ways. Sometimes you will see optional<ref<T>> instead of weak_ptr, this is because it is more correct and nicer to work with in practice.

2. The base UML diagram shows practically nothing on which values HAVE to be mutable and why. Ownership does not show, so I implemented it to the best of my ability and left TODO markers where I was not sure (see gameObject.h).

3. Changed the name of IManager to IEngineService because of the naming. IManager is just way to verbose and abstract.

### DIFF
--- a/include/engine/core/engine.h
+++ b/include/engine/core/engine.h
@@ -1,11 +1,16 @@
 #pragma once
+#include <memory>
 
-#ifndef ENGINE_CORE_ENGINE_H
-#define ENGINE_CORE_ENGINE_H
+#include <engine/core/serviceContainer.h>
 
 class Engine {
+private:
+    Engine();
 public:
-    Engine(); // Temporary
-};
+    static Engine& instance() {
+        static Engine engine {};
+        return engine;
+    }
 
-#endif // ENGINE_CORE_ENGINE_H
+    const std::unique_ptr<ServiceContainer> services;
+};
--- a/include/engine/public/component.h
+++ b/include/engine/public/component.h
@@ -9,13 +9,17 @@
  * If it is not used for the same reason, make the naming more specific to avoid confusion.
  * I could not tell what it was for, so I left it out for now.
  */
-class Component : public GameObject {
+class Component {
 private:
-    std::weak_ptr<GameObject> _parent;
+    std::optional<std::reference_wrapper<GameObject>> parent_;
+    bool active_ = true;
     // TODO: Maybe add the active property back with a more specific name if needed.
 public:
-    explicit Component(const Scene& scene);
-    ~Component() override = default;
+    explicit Component(GameObject& parent);
+    virtual ~Component() = default;
+
+    [[nodiscard]] bool active() const noexcept;
+    Component& active(bool value) noexcept;
 
     virtual void update() = 0;
     virtual void onAttach() = 0;
--- a/include/engine/public/component.h
+++ b/include/engine/public/component.h
@@ -1,7 +1,8 @@
 #pragma once
 #include <memory>
+#include <optional>
 
-#include <engine/public/gameObject.h>
+class GameObject;
 
 /**
  * @brief Base class for all components that can be attached to GameObjects.
@@ -29,4 +30,4 @@ public:
 };
 
 template<typename T>
-    concept IsComponent = std::is_base_of<Component, T>::value;
+    concept IsComponent = std::is_base_of_v<Component, T>;
--- a/include/engine/public/component.h
+++ b/include/engine/public/component.h
@@ -25,4 +25,4 @@ public:
 };
 
 template<typename T>
-    concept IsComponent = std::is_base_of<Component, T>::type;
+    concept IsComponent = std::is_base_of<Component, T>::value;
--- a/include/engine/public/component.h
+++ b/include/engine/public/component.h
@@ -4,17 +4,10 @@
 
 class GameObject;
 
-/**
- * @brief Base class for all components that can be attached to GameObjects.
- * I left out the active: bool property because the GameObject already has an active property.
- * If it is not used for the same reason, make the naming more specific to avoid confusion.
- * I could not tell what it was for, so I left it out for now.
- */
 class Component {
 private:
     std::optional<std::reference_wrapper<GameObject>> parent_;
     bool active_ = true;
-    // TODO: Maybe add the active property back with a more specific name if needed.
 public:
     explicit Component(GameObject& parent);
     virtual ~Component() = default;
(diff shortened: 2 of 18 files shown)


ThomasRuijters:
I would like to discuss some of the TODO points and the usage of weak pointers. I found exactly _one_ case where we used them correctly. They are not safe nullable references, they are not meant for that. They represent permanent non-owning relationships where the subject could potentially vanish. It does not represent a value that can change or could be null in a relationship. They exist for a very specific reason and that reason does not really apply to our code base at the moment nor does it to our current model. It **could** be done and **technically** works, but its ugly. That it works does not mean it works properly, and with weak_ptr instances everywhere we are communicating the wrong semantic meaning and poorly express intent. Keep it to optional<ref<T>> unless the relationship is terminating if broken. weakptr does not mean that you should keep mutating the things they are meant to be set once. If you want ownership semantics just transfer. Bad usage, do not do it. I'd prefer to remove them from GameObject and Component entirely.

include/engine/public/transform.h
```diff
@@ -0,0 +1,31 @@
...
+    std::optional<std::reference_wrapper<Transform>> parent_;
+public:
+
+    Transform();
+    explicit Transform(Vector3 position);
+    Transform(Vector3 position, float rotation, int scale);
+    Transform(Vector3 position, float rotation, int scale, std::optional<std::reference_wrapper<Transform>> parent);
+
+    Transform& position(const Vector3& pos) noexcept;
+    [[nodiscard]] Vector3 position() const noexcept;
```
BotSquirrel: I am missing a setter and getter for `localposition`. Once we look at transforms with parents `position` will return its current position + that of the parent. We still need a way then to update and access that without the link, hence a local variant

include/engine/public/gameObject.h
```diff
@@ -0,0 +1,112 @@
...
+    Alternative: Make it shared so multiple objects can share components.
+    It would be nice if you want to do complex shared behaviors but its complex, don't. */
+    std::vector<std::unique_ptr<Component>> components_ {};
+
+    /** GameObject does not own its children. Why? Because every GameObject instance is owned by a scene.
+        If you turn this around, a Scene would only own root objects. That works, but it isn't what we
+        want to happen. A shared_ptr would also not work because you would get weird undefined behaviors.
+        In this case its paramount that a GameObject is destroyed with the intention to keep its
+        children alive, or to have them be destroyed all together on destruction. */
+    std::vector<std::weak_ptr<GameObject>> children_ {};
```
BotSquirrel: If this proves to be challenging then we could always look at scene. Filter the components based on parent and return that rerefence? Its just a suggestion "if", the current way is fine

include/engine/public/component.h
```diff
@@ -0,0 +1,28 @@
...
+ * I left out the active: bool property because the GameObject already has an active property.
+ * If it is not used for the same reason, make the naming more specific to avoid confusion.
+ * I could not tell what it was for, so I left it out for now.
+ */
+class Component : public GameObject {
+private:
+    std::weak_ptr<GameObject> _parent;
+    // TODO: Maybe add the active property back with a more specific name if needed.
+public:
+    explicit Component(const Scene& scene);
```
BotSquirrel: We could pass the scene upon creation or fetch from our parent? As without a gameobject what value has a component?

include/engine/public/gameObject.h
```diff
@@ -0,0 +1,112 @@
...
+    std::optional<std::reference_wrapper<GameObject>> parent_;
+    bool isActive_ {true};
+
+public:
+    // TODO: Could we for the love of god make some of these immutable and private? I do not know if these have to be public or mutable right now...
+    std::string name;
+    std::string tag;
+    int layer {};
+    Scene& scene;
+    Transform transform;
```
BotSquirrel: In my opinion all of the member variables could be protected with public accessors: `name()`, `name(const std::string& n)`, etc. Pretty much all of these func's can be public as we need to access them. I don't think gameobject needs that much private classes as it forms a base for others.

src/public/gameObject.cpp
```diff
@@ -0,0 +1,70 @@
...
+GameObject::~GameObject() {
+    if (parent_.has_value()) {
+        parent_->get().removeChild(*this);
+    }
+}
+
+
+// TODO: Discuss whether this should be recursive (i.e., check parents' active states)
+// TODO: Discuss if a GameObject its active state should be determined by its parent
+bool GameObject::isActiveInWorld() const noexcept {
```
BotSquirrel: If I take a peek at unity, then it would not be recursive. Its a simple toggle that only looks at itself. Ideally this also is not used that much. If we need to hide something then why would i exist? If its a logic object for a script then do not give it a rigidbody and sprite.

include/engine/public/gameObject.h
ThomasRuijters: On it

include/engine/core/engine.h
```diff
@@ -1,11 +1,16 @@
...
 public:
-    Engine(); // Temporary
-};
+    static Engine& instance() {
+        static Engine engine {};
+        return engine;
+    }
 
-#endif // ENGINE_CORE_ENGINE_H
+    const std::unique_ptr<ServiceContainer> services;
```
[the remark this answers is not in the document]
ThomasRuijters: This is supposed to be a mutable value, its just not reassignable

src/public/gameObject.cpp
```diff
@@ -0,0 +1,104 @@
+#include <engine/public/gameObject.h>
+#include <engine/public/component.h>
+
+GameObject::GameObject(Scene& scene) : scene_(scene) {}
+GameObject::~GameObject() {
+    if (parent_.has_value()) {
+        parent_->get().remove_child(*this);
+    }
+}
```
BotSquirrel, line 9: It would be nice if somebody *later* could alter this so when a scene gets added it automatically adds itself to the scene.

BotSquirrel (approved):
My comment is more something for future use I would like to see. I see no warnings or issues and expect it not to leak. @ThomasRuijters and I talked about the concerns of using `reference_wrappers`, I personally think that we will see how it goes. If it works and everybody can use it, fine by me, if not then we need to refactor a small bit. As we are still in the early stages that should not be that big of a problem

src/public/util/point.cpp
```diff
@@ -0,0 +1,4 @@
+#include <engine/public/util/point.h>
+
+Point::Point() : x(0.0f), y(0.0f) {}
+Point::Point(const float x, const float y) : x(x), y(y) {}
```
LarsDevans, line 4: Voor primitieve typen heeft 'const' geen effect

LarsDevans (approved):

PR #20, opened 2025-10-27
feat (ci-tests): add ci/cd with a basic test example

BotSquirrel:
> [!NOTE]
> This is just the CI/CD pipeline with a very basic (GoogleTest) test that contains no logic. The engine library is linked with the tests files. We might want to discuss if we group the tests under a single command as i do now, or seperate them.
> 
> The reason I decided to use GoogleTests is because its easy, simple and feels familiair to the way we know and have handled Unit tests

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,10 +1,5 @@
 name: C++ CI
 on:
-  push:
-    branches:
-      - master
-      - development
-      - '@feat/*'
   pull_request:
     branches:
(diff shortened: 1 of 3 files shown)


ThomasRuijters (approved):
Runs and compiles, so I see no problem. May be a bit of a wheee slow guy, but as long as we only do it for PR's to main and dev its fine.

LarsDevans (approved):

PR #21, opened 2025-10-27
feat (profiling): add tracy profiler client and memory leak detection

BotSquirrel:
> [!IMPORTANT]
> Tracy is installed as a client, meaning you need to run the `profiler` from the latest [release](https://github.com/wolfpld/tracy/releases/tag/v0.12.2). To launch it:
> 1. Extract the `profiler` from the release
> 2. Run the profiler
> 3. Start the engine
> 4. In the profiler connect with the application
> 5. Inspect and profit
> 6. Close the engine and see if you leaked something
>
> You do not need to run the profiler to check for memory leaks, that happens automatically and is dumped in the console. I also added an option to disable it for the pipeline checks.
> ***Side note for @LarsDevans***, I am not sure if the profiler runs on linux/mac... sorry for that in advance. However, you should still see the memory leaks.

### DIFF
--- a/include/engine/util/memory.h
+++ b/include/engine/util/memory.h
@@ -1,8 +1,5 @@
 #pragma once
 
-#ifndef MEMORY_H
-#define MEMORY_H
-
 #include <cstddef>
 
 void tracy_dump_leaks();
@@ -14,5 +11,3 @@ void* operator new(std::size_t size);
 void operator delete(void* ptr) noexcept;
 void* operator new[](std::size_t size);
 void operator delete[](void* ptr) noexcept;
-
-#endif // MEMORY_H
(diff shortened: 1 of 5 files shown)


ThomasRuijters (approved):
"It works on my machine"

include/engine/util/memory.h
```diff
@@ -0,0 +1,18 @@
+#pragma once
+
+#ifndef MEMORY_H
+#define MEMORY_H
```
LarsDevans: One is enough :) I prefer '#pragma once'

LarsDevans (approved):
> Side note for @LarsDevans, I am not sure if the profiler runs on linux/mac... sorry for that in advance. However, you should still see the memory leaks.

I don't think it's suited for Linux and macOS:
image

I got my own tooling; so no worries.

Issue #22, opened 2025-10-27
Input manager skeleton


LarsDevans:
### Summary

Create the skeleton implementation of the `InputManager` system, responsible for handling keyboard input within the engine. This issue focuses on architecture and interface design, no concrete input backend (e.g., SDL, GLFW, etc.) should be implemented yet.

### Goals

- Define the `InputManager` class and register it with the **Engine Services Registry** so it can be globally accessed.
- Create an `InputProvider` interface or abstract base to expose methods for key state queries:
    - `is_key_down(KeyCode key)`
    - `is_key_up(KeyCode key)`
- Define a `KeyCode` enum that includes all standard keyboard keys.
- Implement a `KeyState` structure or mechanism to track key transitions (pressed, held, released) across frames.

### Deliverables

- Header and stub implementation files for:
    - `InputManager`
    - `InputProvider`
- `KeyCode` and `KeyState`
- Registration of the input manager within the engine’s service registry.
- Documentation/comments outlining how concrete input backends will integrate later.

LarsDevans:
The `InputManager` registration to the engine is not yet possible, because the registration mechanism is not yet implemented.

PR #29, opened 2025-10-28
feat(input): add skeletons

LarsDevans:
See #22 for information about the task

### DIFF
--- a/include/engine/input/i_input_provider.h
+++ b/include/engine/input/i_input_provider.h
@@ -2,7 +2,7 @@
 
 #include <optional>
 
-#include "engine/input/key_code.h"
+#include <engine/input/key_code.h>
 
 /**
  * @brief Interface for querying low-level keyboard input.
--- a/include/engine/input/i_input_provider.h
+++ b/include/engine/input/i_input_provider.h
@@ -1,0 +1,86 @@
+#pragma once
+
+#include <optional>
+
+#include "engine/input/key_code.h"
+
+/**
+ * @brief Interface for querying low-level keyboard input.
+ *
+ * Defines a unified polling API for backend-specific input providers
+ * (e.g., SDL, GLFW, or platform-native implementations).
+ */
+class IInputProvider {
+public:
+    virtual ~IInputProvider() = default;
+
+    /**
+     * @brief Returns whether the specified key is currently held down.
+     *
+     * Suitable for continuous input such as movement or sustained actions.
+     */
+    virtual bool is_key_held(KeyCode key) const = 0;
+
+    /**
+     * @brief Returns true on the frame the specified key transitions to pressed.
+     *
+     * Supports one-shot actions such as jumping or confirming selections.
+     */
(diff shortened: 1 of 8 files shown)


include/engine/input/i_input_provider.h
```diff
@@ -0,0 +1,86 @@
+#pragma once
+
+#include <optional>
+
+#include "engine/input/key_code.h"
```
BotSquirrel: Because we built the engine as a seperate library we can `#include <engine/input/key_code.h>`. Works the same but might be cleaner and consisten to use the `<>` variant?

include/engine/input/i_input_provider.h
```diff
@@ -0,0 +1,86 @@
...
+
+/**
+ * @brief Interface for querying low-level keyboard input.
+ *
+ * Defines a unified polling API for backend-specific input providers
+ * (e.g., SDL, GLFW, or platform-native implementations).
+ */
+class IInputProvider {
```
BotSquirrel, line 13: So what do we do with mouse inputs? `mouse_left_click`,  `mouse_right_click` and `mouse_middle_click` could be interpreted als buttons. I get that we might want to write something extra for mouse movement as we need the `axis`, or we could add something here as technically its input. If we write something outside of this class for mouse input then it should something like a `IKeyInputProvider`

include/engine/input/input_manager.h
```diff
@@ -0,0 +1,15 @@
...
+/**
+ * @brief Centralized input manager that delegates to an external input provider.
+ */
+class InputManager {
+public:
+    IInputProvider* provider() const noexcept;
+    void set_provider(IInputProvider* provider) noexcept;
+
+private:
+    IInputProvider* _provider = nullptr;
```
BotSquirrel: Raw pointer 👻? Unique pointer would work as this class is the owner and the provider is never transfered?

include/engine/input/input_manager.h
```diff
@@ -0,0 +1,15 @@
+#pragma once
+
+#include "engine/input/i_input_provider.h"
+
+/**
+ * @brief Centralized input manager that delegates to an external input provider.
+ */
+class InputManager {
```
BotSquirrel: No inheritence from `iEngineService`? I know in the diagram we wrote `IManager`, its practically the same just has a different name

include/engine/input/i_input_provider.h
LarsDevans: Let's discuss this when we meet

include/engine/input/input_manager.h
LarsDevans: That did not exist when I started this PR. Will change this now

BotSquirrel (approved):
Small PR's right... 😉 I see it got bigger due to the `@feat` not being up-to-date with `development`. All my comments, except the mouse thing, is fixed so its got my blessing

SjulGr (approved):

PR #32, opened 2025-10-28
feat: SDL_Wrappers

ThomasRuijters:
Add SDL wrappers for Renderer and Texture according to the UML diagram.

### DIFF
--- a/src/core/rendering/Renderer.cpp
+++ b/src/core/rendering/Renderer.cpp
@@ -1,0 +1,59 @@
+#include <engine/core/rendering/Renderer.h>
+#include <engine/public/gameObject.h>
+#include <SDL3_image/SDL_image.h>
+
+#include "engine/core/rendering/Texture.h"
+
+Renderer::Renderer(SDL_Renderer* renderer)
+    : _renderer(std::unique_ptr<SDL_Renderer, decltype(&SDL_DestroyRenderer)>{renderer, SDL_DestroyRenderer}) {
+    SDL_SetRenderDrawColor(_renderer.get(), 0, 0, 0, 255);
+}
+Renderer::Renderer(SdlRendererPtr renderer)
+    : _renderer(std::move(renderer)) {
+}
+
+void Renderer::clear() const {
+    SDL_RenderClear(_renderer.get());
+}
+void Renderer::render() const {
+    SDL_RenderPresent(_renderer.get());
+}
+
+// TODO: Remove dummy when implementing Sprite component, and implement the RenderObject method accordingly...
+class Sprite : public Component {
+    public:
+    Texture* texture;
+    explicit Sprite(GameObject& parent) : Component(parent) {
+
+    }
+};
+
+// TODO: Reimplement this when implementing the renderer
+void Renderer::renderObject(GameObject& gameObj) const{
+    const auto& transform = gameObj.transform();
+    const auto& absolutePosition = transform.absolute_position();
+
+    for (auto spriteWrapper : gameObj.get_components<Sprite>()) {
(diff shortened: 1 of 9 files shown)


src/core/rendering/renderer.cpp
```diff
@@ -0,0 +1,65 @@
...
+}
+
+// TODO: Remove dummy when implementing Sprite component, and implement the RenderObject method accordingly...
+class Sprite : public Component {
+    public:
+    Texture* texture{};
+    explicit Sprite(GameObject& parent) : Component(parent) {
+
+    }
+};
```
BotSquirrel, line 32: be sure to delete it later or pull out of this file and create own class file?

src/core/rendering/renderer.cpp
```diff
@@ -0,0 +1,65 @@
...
+
+// TODO: Reimplement this when implementing the renderer
+void Renderer::render(const std::vector<std::reference_wrapper<GameObject>>& objects) const{
+    for (auto gameObjWrapper : objects) {
+        auto& gameObj = gameObjWrapper.get();
+        const auto& transform = gameObj.transform();
+        const auto& absolutePosition = transform.absolute_position();
```
BotSquirrel, line 39: If i'm not mistaken then this should always be `position` instead as we want to draw it relative to its parent. This would draw it relative to the scene, so a `box1` with a `1,1` and a child `box2` with `1,1` would return `1.1` for the child while that would be `2,2`

src/core/rendering/Renderer.cpp
```diff
@@ -0,0 +1,65 @@
...
+                .y = absolutePosition.y,
+                .w = texture.width(),
+                .h = texture.height()
+            };
+
+            SDL_RenderTextureTiled(
+                renderer_.get(),
+                sprite.texture->texture_.get(),
+                &source,
+                1,
```
BotSquirrel: What is this `1` you speak of? Pull into variable or comment?

BotSquirrel (changes requested):

src/core/rendering/renderer.cpp
ThomasRuijters: For now its just there as a placeholder. We need something for it and to put it in its own file is just a step to far. Its basically a fleshed out forward that is all. If you do not like it its one of those to bad cases. We need it for now, it will be removed on the next merge. Does that satisfy for now?

src/core/rendering/renderer.cpp
ThomasRuijters: SDL does not care about nesting. SDL takes a rect and draws it somewhere. The absolute position of something is just the position of that thing compared to the world, not its parent. That is what we need. If you draw everything relative to its parent you start shooting yourself in the foor. Its a logical dependency we do not need and thus should not have. Do not rely on technical implementation to dictate semantics.

BotSquirrel: Sure you do you, and i suppose we'll see when someone draws something. It is still my guess that this will draw things in the wrong position.

src/core/rendering/renderer.cpp
BotSquirrel: Just thought to highlight it and raise awareness, as i've said if its deleted later when there is an implementation then its fine 👌

src/core/rendering/renderer.cpp
ThomasRuijters: We could have a talk about it for sure. I think it is worth discussing. To me, relative positions are positions compared to the parent. Absolute positions are basically the exact location of their parent + their own relative position. This bubbles up basically.

BotSquirrel (approved):

PR #38, opened 2025-11-03
feat: add rendering service

ThomasRuijters:
Do we want to keep the renderingservice class? All it does is wrap the renderer which wraps SDL and forward a call. I vote we rename the renderer to renderingService and be done with it.


BotSquirrel (approved):
Nothing out of the ordinary, the tests fail but thats on me (@BotSquirrel). I hate not being able to run my pipelines locally. I'll push another fix PR...

SjulGr (approved):

Issue #33, opened 2025-10-29
Audio service implementation


BotSquirrel:
### Summary

Fill the skeleton for the `audio service` with a working implementation.

### Goals

- Be able to register a sound instance
- Be able to start playing a sound instance
- Be able to stop a sound instance
- Be able to get a sound
- Be able to get all sounds

### Deliverables

- A working `audio service` that can be used by the service container

PR #40, opened 2025-11-04
feat (audio): implement sound instances and resources

BotSquirrel:
See #33 for more information about the task.

During the implementation I found out that the current setup would not work as I wanted, I could not save a resource and create copies to play. That way a sound has one base and multiple players instead of duping everything. hence, I refactored the instances and split them into resources. The PR did increase in size due to that.

Before anyone tags me about tests, I wanted to but it proved quite difficult and is something I want to discuss. If someone still wants to try it see below. The actual code is 3 lines except you need an SDL instance for audio...

```cpp
// main.cpp
void run()
{
    // mini audio test with very basic SDL window
    SDL_Init(SDL_INIT_VIDEO | SDL_INIT_AUDIO);
    SDL_Window* window = SDL_CreateWindow("Audio Service Test", 800, 600, SDL_WINDOW_RESIZABLE);
    SDL_Renderer* renderer = SDL_CreateRenderer(window, nullptr);

    AudioService audioService;
    std::string filePath = "./resources/sounds/bonetrousle.wav";
    std::string soundName = "Bonetrousle";

    auto soundResource = audioService.register_sound(filePath, soundName, SoundType::SDL_MIXER);
    auto soundInstance = audioService.play_sound(soundName, 0.1f, true);
    
    SDL_Event e;
    bool running = true;
    while (running) {
        while (SDL_PollEvent(&e))
        {
            if (e.type == SDL_EVENT_QUIT || (e.type == SDL_EVENT_KEY_DOWN && e.key.key == SDLK_ESCAPE)) {
                running = false;
                break;
            }
            if (e.type == SDL_EVENT_KEY_DOWN && e.key.key == SDLK_P) {
                audioService.play_sound(soundName, 1.0f);
                std::cout << "Playing sound: " << soundName << std::endl;
            }
            if (e.type == SDL_EVENT_KEY_DOWN && e.key.key == SDLK_S) {
                audioService.stop_all_sounds();
                std::cout << "Stopped all sounds." << std::endl;
            }
            if (e.type == SDL_EVENT_KEY_DOWN && e.key.key == SDLK_O) {
                for (const auto& instance : audioService.get_all_playing_sounds()) {
                    std::cout << "Currently playing sound: " << instance->resource()->name() << std::endl;
                }
                break;
            }

            audioService.update();
            SDL_Delay(16);
        }
    }
}
```


include/engine/audio/audio_service.h
```diff
@@ -0,0 +1,48 @@
...
+#include <engine/audio/sound/resource/sound_resource.h>
+#include <engine/audio/sound/sound_type.h>
+
+/**
+ * @brief Manages audio playback and sound instances within the engine.
+ */
+class AudioService : public IEngineService {
+public:
+    AudioService() = default;
+    ~AudioService() override = default;
```
ThomasRuijters: We do not really do anything with the destructor here. The basis is already default (it cleans just its present members) so I don't know if an override is needed here.

include/engine/audio/audio_service.h
```diff
@@ -0,0 +1,48 @@
...
+
+    std::shared_ptr<SoundResource> register_sound(const std::string& file_path, const std::string& name, SoundType type = SoundType::GENERIC);
```
ThomasRuijters, line 21: How come we have a shared pointer here? We rarely ever need those? Is it not possible to assign a single resource owner here?

include/engine/audio/audio_service.h
```diff
@@ -0,0 +1,48 @@
...
+
+    void stop_sound(std::shared_ptr<SoundInstance>&& sound_instance);
+    void stop_sound(const std::string& name);
+    void stop_all_sounds();
+
+    std::shared_ptr<SoundInstance> get_sound_instance(const std::string& name) const noexcept;
+    std::vector<std::shared_ptr<SoundInstance>> get_all_playing_sounds() const noexcept;
+
+    void set_sound_volume(std::shared_ptr<SoundInstance> sound_instance, float volume) noexcept;
+    void set_sound_volume(const std::string& name, float volume) noexcept;
```
ThomasRuijters, line 38: Isn't it better to make the soundInstance have these member functions instead? Make the sound instance refer back to the manager if need be. This looks like we are controlling the property of a SoundInstance object from a manager. Could do, but unless that property needs to be manager by the manager I would recommend a move here.

include/engine/audio/sound/instance/generic_sound_instance.h
ThomasRuijters, line 1: Are we ever going to use this? I get that we have one, but we have one for SDL already. Feels like overkill but its nice that we have it.

include/engine/audio/sound/sound_factory.h
```diff
@@ -0,0 +1,22 @@
...
+
+/**
+ * @brief Factory class for creating SoundResource instances based on SoundType.
+ * The reason for shared pointers is to manage the lifetime of sound resources and instances
+ * easily share them across different parts of the audio system without worrying about ownership.
+ */
+class SoundFactory {
+public:
+    SoundFactory() = default;
+    ~SoundFactory() = default;
```
ThomasRuijters: If the class is static or abstract just make them both 0. That way it truely becomes a wrapper. Either that or make it a singleton. I vote static class and be done with it. I believe that if you just put functions in a namespace you do not even need the class, its just grouping at this point.

ThomasRuijters (commented):
In general I have very little issues with the code at hand. Do mind that my eyes twitch when I see shared_ptr instances...

src/audio/sound/instance/generic_sound_instance.cpp
```diff
@@ -0,0 +1,36 @@
...
+}
+
+void GenericSoundInstance::stop() {
+    playing_ = false;
+    paused_ = false;
+    finished_ = true;
+}
+
+bool GenericSoundInstance::is_playing() const noexcept {
+    return playing_ && !paused_;
```
ThomasRuijters, line 27: Shouldn't we just set playing to false if we pause the thing? Do we still have to check both? In this case playing is the source of truth, not paused...

src/audio/sound/resource/sdl_sound_resource.cpp
```diff
@@ -0,0 +1,44 @@
...
+
+    if (file_path.size() < 4 || file_path.substr(file_path.size() - 4) != ".wav") {
+        throw std::runtime_error("SDLSoundResource only supports .wav files: " + file_path);
+    }
+
+    if (!std::filesystem::exists(file_path)) {
+        throw std::runtime_error("WAV file does not exist: " + file_path);
+    }
+    
+    try {
```
ThomasRuijters, line 21: then why even catch it?

ThomasRuijters (approved):
I do have some nitpicks but the overall code looks good.

include/engine/audio/audio_service.h
BotSquirrel: So I agree on some part of this. Resources are shared across instances, as each instance knows about a resource. I could have made it a reference perhaps but this ecosystem runs on its own basically. When looking at the usage:
```cpp
auto soundResource = audioService.register_sound(filePath, soundName, SoundType::SDL_MIXER);
auto soundInstance = audioService.play_sound(soundName, 0.1f, true);
```
We get a shared_pointer and create an instance. Then when we leave it gets cleanup up here but stays in the instance.

On the part where i agreed, instances can be `unique_ptr` as they are clones but unique ones. So i made them so using reference wrappers.

include/engine/audio/audio_service.h
BotSquirrel: The `SoundInstance` already has those, but ideally you interact with the service rather then the instance. It simply delegates it. Due to SDL limitations we also can't push the audio directly so thats why the service is prefered

include/engine/audio/sound/instance/generic_sound_instance.h
BotSquirrel: Perhaps not but I build it dynamically and needed a default as otherwise it would not compile. Can't instantiate an abstract class which is `SoundInstance`.

src/audio/sound/instance/generic_sound_instance.cpp
BotSquirrel: We already do that (playing will be false and paused is true on pause). This is more of an extra safety check to make sure

src/audio/sound/resource/sdl_sound_resource.cpp
BotSquirrel: To avoid a breakdown of the Engine? I suppose we wouldn't want the entire thing to crash because a `.wav` file was moved

SjulGr (approved):

Issue #34, opened 2025-10-29
Input service for mouse research


LarsDevans:
### Summary

Design and define the mouse input system architecture, extending the existing `InputService` framework. The goal is to establish the interfaces, data structures, and registration points for mouse input, without binding to any specific platform or input library yet.

### Goals

- Determine where mouse input should be managed:
    - Evaluate whether mouse handling belongs inside the existing `InputService` or as a separate `MouseInputProvider` registered to it.
    - The preferred design should allow `InputService` to query or delegate mouse events to specialized providers.
- Define methods and structures for common mouse interactions:
    - `is_mouse_button_down(MouseButton button)`
    - `is_mouse_button_up(MouseButton button)`
    - `get_mouse_position()` (returns current cursor coordinates)
    - `get_mouse_delta()` (movement since last frame)
    - Optionally prepare for `get_scroll_delta()` (for scroll wheel events)
- Define `MouseButton` enum (Left, Right, Middle, etc.)
- Create a `MouseState` struct/class to track button transitions and position changes between frames.

### Deliverables

- Header and stub implementation files for:
    - `MouseInputProvider` (interface or base class)
    - `MouseState` and `MouseButton`
- Integration points in the `InputService` (if applicable):
    - Registration and update calls
    - Public accessors for mouse queries
- Documentation/comments

PR #41, opened 2025-11-05
feat (input): add mouse skeletons

LarsDevans:
See #34  for information about the task


BotSquirrel (approved):
Nice I think that covers all needed mouse actions, we're not making an RPG so I get why we are only mapping 5 keybtn's. 
Only nitpicky thing I could see is that perhaps all `float` pairs should be a `vector2` once it exists? Now there is a bit of a mixture while the `Point`'s could also be a `vector2` to keep it all consistent

LarsDevans:
@BotSquirrel about the point/vector comment; there's a mathematical difference between the two. I've used the point datatype where it is applicable, given the context of it. I _can_ change the floats the points, totally doable. Perhaps it's something to discuss next meeting

SjulGr (approved):

PR #47, opened 2025-11-07
feat(input): implement manager

LarsDevans:
The signature of the provider's getter is now marked as const. The signature could change depending on the function signatures of SDL's implementation that will be used

### DIFF
--- a/src/input/input_manager.cpp
+++ b/src/input/input_manager.cpp
@@ -1,0 +1,14 @@
+#include <engine/input/input_manager.h>
+#include <engine/input/i_input_provider.h>
+
+const IInputProvider& InputManager::provider() const
+{
+    if (!provider_)
+        throw std::runtime_error("Input provider is not set.");
+    return *provider_;
+}
+
+void InputManager::set_provider(std::unique_ptr<IInputProvider> provider) noexcept
+{
+    provider_ = std::move(provider);
+}


BotSquirrel (approved):

ThomasRuijters (changes requested):
Getters and setters are written as: prop(): value and prop(value): this. So write it as:

```cpp
IInputProvider& InputManager::provider(std::unique_ptr<IInputProvider> provider) noexcept
{
    provider_ = std::move(provider);
    return *this;
}
```
Do mind that you are passing a unique pointer here. Are we transferring ownership or are we acquired a resource that has no owner yet? Be explicit in what is happening here. Either way the unique_ptr is fine just make sure you know what it is you are saying here.

LarsDevans:
@ThomasRuijters the `InputManager` is taking ownership of whatever input provider you pass to it. I don't understand the benefit of changing the method's signature to the one you suggest. Right now the getter and setter are nicely separated. With the one you provide, the method is changing state and returning something. That might be confusing, no?

ThomasRuijters:
> @ThomasRuijters the `InputManager` is taking ownership of whatever input provider you pass to it. I don't understand the benefit of changing the method's signature to the one you suggest. Right now the getter and setter are nicely separated. With the one you provide, the method is changing state and returning something. That might be confusing, no?

I get the unique ptr being passed, it was but an observation nothing to worry about. The getter setter syntax is something that we agreed on. The reason why you usually want to return the source object is due to semantics. I can brief you on it later on, but it comes down to one being a side effect and the other (a read) being a pull of information from a source. Returning the source makes it clear that you are performing an effect and are returning the result of that effect, being the source itself.

LarsDevans:
@ThomasRuijters let's get this PR merged tomorrow. We'll discuss this during or after the check-in

LarsDevans:
It's good to go

PR #49, opened 2025-11-07
feat(input): implement mouse state with tests

LarsDevans:


BotSquirrel (approved):

tests/input/mouse_state.cpp
```diff
@@ -0,0 +1,78 @@
+#include <catch2/catch_test_macros.hpp>
+
+#include <engine/input/mouse_state.h>
+
+TEST_CASE("mouse_button_state_reset_should_set_current_and_previous_to_false", "[MouseButtonState]") {
+    MouseButtonState buttonState;
```
BotSquirrel: camelCase but other variables snake_case? (I made the same mistake). Personally I am not that nitpicky on tests so I approve it but its good to remember

SjulGr (approved):

PR #59, opened 2025-11-10
feat (audio-service): merge the final service to development

BotSquirrel:
see #33 for further details. TLDR; the entire audio service including tests


include/engine/audio/audio_service.h
```diff
@@ -0,0 +1,45 @@
...
+#include <engine/audio/sound/sound_type.h>
+
+/**
+ * @brief Manages audio playback and sound instances within the engine.
+ */
+class AudioService : public IEngineService {
+public:
+    AudioService() = default;
+
+    std::shared_ptr<SoundResource> register_sound(const std::string& file_path, const std::string& name, SoundType type = SoundType::GENERIC);
```
ThomasRuijters: Not a fan of shared pointers. This will dictate that everything has to use them. Consider ref wrappers instead with unique pointers. Not a must, nice to have.

include/engine/audio/audio_service.h
ThomasRuijters, line 1: This thing looks so much like the resource pool for textures that we may even be able to fuse them together. One abstraction called IResourcePool<T> and the rest is magic.

include/engine/audio/sound/sound_factory.h
```diff
@@ -0,0 +1,22 @@
...
+ * The reason for shared pointers is to manage the lifetime of sound resources and instances
+ * easily share them across different parts of the audio system without worrying about ownership.
+ */
+class SoundFactory {
+public:
+    SoundFactory() = delete;
+    ~SoundFactory() = delete;
+
+    static std::shared_ptr<SoundResource> create_sound_resource(const std::string& file_path, const std::string& name, SoundType type = SoundType::GENERIC);
+    static std::unique_ptr<SoundInstance> create_sound_instance(std::shared_ptr<SoundResource>& resource, float volume = 1.0f);
```
ThomasRuijters, line 21: Why shared for one and unique for the other?

src/audio/audio_service.cpp
```diff
@@ -0,0 +1,169 @@
...
+        return it->second;
+    }
+
+    try {
+        auto resource = SoundFactory::create_sound_resource(file_path, name, type);
+        sound_resources_[name] = resource;
+
+        return resource;
+    } catch (...) {
+        return nullptr;
```
ThomasRuijters: watch out with returning nullptr here. Use option<T> if you are not sure that the thing is going to have a value.

src/audio/audio_service.cpp
```diff
@@ -0,0 +1,169 @@
...
+    auto sound_resource = get_sound_resource(name);
+
+    if (!sound_resource) {
+        throw std::invalid_argument("Sound resource not found: " + name);
+    }
+
+    return play_sound(sound_resource, volume, loop);
+}
+
+void AudioService::stop_sound(std::unique_ptr<SoundInstance>&& sound_instance) {
```
ThomasRuijters, line 78: Why is the value destructive?

ThomasRuijters (changes requested):
I made some comments that are valid in more then one place. You will know what to do with them when you see them.

include/engine/audio/audio_service.h
BotSquirrel: I already explained this in the previous PR on the same topic. TLDR; resources (audio data) are shared across multiple instances. It could be a reference wrapper but its meant to be shared and behave like that. This way when an instance stops (a song ends) its deleted without deleting the resource without any extra complications. Besides we never use it outside and is only meant for inside this service. As it never leaves it can't dictate other things

include/engine/audio/audio_service.h
BotSquirrel: Yes thats why initiality during the diagram phase I suggested this already but we pulled it apart for convenience. Now that we all are developing I think its prone to less issues if we keep it this way

include/engine/audio/sound/sound_factory.h
BotSquirrel: Look at the top comment about why I've decided to use shared pointers

src/audio/audio_service.cpp
BotSquirrel: Makes it easier to utilize the function. We get a pointer from playing and then destroy it. Ideally you never call this but there can always be something that abruptly ends things

ThomasRuijters (approved):

SjulGr (approved):

PR #65, opened 2025-11-11
feat: implement basis for sprite component

ThomasRuijters:


include/engine/core/assetManager.h
```diff
@@ -18,18 +18,28 @@ class AssetManager : public IEngineService{
...
+        std::reference_wrapper<const std::vector<std::reference_wrapper<Texture>>>,
+        std::string> create_spritesheet_for(const std::string& source, const std::string& name, size_t from, size_t to);
+
+    [[nodiscard]]
+    std::optional<std::reference_wrapper<Texture>> try_get_texture(const std::string& sprite) const;
+
+    [[nodiscard]]
+    std::expected<std::reference_wrapper<Texture>, std::string> register_texture(
+        const std::string& resourceName,
+        const std::string& textureName,
```
BotSquirrel: `snake_case` 🐍

include/engine/public/components/sprite.h
```diff
@@ -0,0 +1,36 @@
...
+    int flipY_;
+    int sortingLayer_;
+    int orderingLayer_;
+    Color color_;
+
+public:
+    Sprite(GameObject& parent, const std::string& sprite, Color color, int flipX, int flipY, int sortingLayer, int orderingLayer);
+
+    [[nodiscard]] int flipX() const;
+    Sprite& flipX(int val);
```
BotSquirrel: `snake_case` 🐍

src/core/assetManager.cpp
```diff
@@ -54,62 +54,96 @@ std::expected<
...
-        int rows,
-        int cols) {
+    const std::string& file,
+    const std::string& name,
+    int rows,
+    int cols) {
+
+const std::string filePath = "resources/" + file;
+
+const auto& rendererService = Engine::instance()
```
BotSquirrel: `snake_case` 🐍

src/public/components/sprite.cpp
```diff
@@ -0,0 +1,68 @@
...
+    const auto& service = Engine::instance().services->getService<AssetManager>().get();
+    const auto maybe_texture = service.try_get_texture(sprite);
+    if (!maybe_texture.has_value()) {
+        throw std::runtime_error("Failed to get texture from sprite");
+    }
+    auto const resource = maybe_texture.value();
+    return resource.get();
+}
+
+Sprite::Sprite(GameObject& parent, const std::string& sprite, const Color color, int const flipX, const int flipY, const int sortingLayer, const int orderingLayer)
```
BotSquirrel: Why not list the constructor first? Is more of an aesthetic point as it serves no value

BotSquirrel (changes requested):

src/public/components/sprite.cpp
ThomasRuijters: Because otherwise I have to forward declare the function

BotSquirrel (approved):

SjulGr (approved):

include/engine/public/components/sprite.h
```diff
@@ -0,0 +1,36 @@
+#pragma once
+#include <string>
+#include <engine/public/util/color.h>
+#include <engine/core/rendering/texture.h>
+#include <engine/public/component.h>
+
+class Sprite : public Component {
+private:
+    const Texture& texture_;
```
LarsDevans: Don't we want this to be changeable: it's usage is more flexible that way

include/engine/public/util/color.h
```diff
@@ -0,0 +1,10 @@
+#pragma once
+
+class Color {
+public:
+    int r;
+    int g;
+    int b;
+    int a;
```
LarsDevans: Can be unsigned, or even denoted as unsigned char

include/engine/public/components/sprite.h
ThomasRuijters: I have modeled it after the idea that if you want a new texture, you throw the old one away. But it could be mutable for sure yea.

include/engine/public/util/color.h
ThomasRuijters: agreed

PR #70, opened 2025-11-12
fix: Make the engine destroyable before the loop ends.

ThomasRuijters:
Due to a goof with static memory and singletons not being nice, we implemented a workaround that should manage the singleton from the instance, not the static class alone.

### DIFF
--- a/include/engine/core/engine.h
+++ b/include/engine/core/engine.h
@@ -1,16 +1,22 @@
 #pragma once
+
 #include <memory>
+#include <iostream>
 
 #include <engine/core/serviceContainer.h>
 
 class Engine {
 private:
+    struct Deleter {
+        void operator()(Engine* e) const { delete e; }
+    };
+
     Engine();
+
+    static std::unique_ptr<Engine, Deleter> engine_instance;
 public:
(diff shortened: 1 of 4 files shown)


include/engine/core/engine.h
```diff
@@ -1,16 +1,22 @@
 #pragma once
+
 #include <memory>
+#include <iostream>
```
BotSquirrel: `<iostream>` is an artefact from debugging and could be removed but I am not blocking over just this

BotSquirrel (approved):

include/engine/core/engine.h
```diff
@@ -1,16 +1,22 @@
...
 
 class Engine {
 private:
+    struct Deleter {
+        void operator()(Engine* e) const { delete e; }
+    };
+
     Engine();
+
+    static std::unique_ptr<Engine, Deleter> engine_instance;
```
SjulGr: Coding guidelines; ending private members on an underscore (`engine_instance` -> `engine_instance_`)

SjulGr (approved):

Issue #67, opened 2025-11-12
Input system adapter for SDL


LarsDevans:
**Description:**
Currently, the input handling logic (e.g. `any_key_down`, `mouse_button_down(MouseButton::primary)`, etc.) is implemented directly inside the engine. This creates tight coupling between the engine and the SDL backend.

**Goal:**
Introduce an adapter layer to separate SDL-specific input handling from the engine core. The engine should depend only on an abstract input interface, while the SDL-specific implementation lives in a dedicated adapter module.

**Tasks:**

- [x] Define an `InputAdapter` interface for polling some concrete implementation
- [x] Implement an `SDLInputAdapter` that uses SDL calls internally
- [x] Update the engine to depend on the `InputAdapter` interface rather than SDL directly
- [x] Ensure existing input-related functionality remains intact

**Benefits:**

- Reduces coupling between engine and SDL
- Makes it easier to add alternative input backends in the future
- Improves testability of engine logic

LarsDevans:
This has been changed to a Strategy pattern. The adapter was found to be unfit/overly complicated

PR #71, opened 2025-11-12
feat(input): use sdl for state update strategy

LarsDevans:
See #67

### DIFF
--- a/include/engine/input/i_input.h
+++ b/include/engine/input/i_input.h
@@ -1,0 +1,14 @@
+#pragma once
+
+#include <map>
+
+#include <engine/input/key_code.h>
+#include <engine/input/key_state.h>
+#include <engine/input/mouse_state.h>
+
+class IInput {
+public:
+    virtual ~IInput() = default;
+
+    virtual void update(std::map<KeyCode, KeyState>& key_states, MouseState& mouse_state);
+};
--- a/include/engine/input/i_input.h
+++ b/include/engine/input/i_input.h
@@ -10,5 +10,13 @@ class IInput {
 public:
     virtual ~IInput() = default;
 
+    /**
+    * @brief Updates the state of keyboard and mouse inputs.
+    *
+    * This function is responsible for querying the current input states, including
+    * key presses, key releases, mouse button presses, releases, and mouse movement.
+    * It updates the provided `key_states` and `mouse_state` with the latest input
+    * information for the current frame.
+    */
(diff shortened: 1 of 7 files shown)


BotSquirrel (approved):

include/engine/input/i_input.h
ThomasRuijters, line 1: the i is part of the name. so no i_Name => instead it is always iName. Its the only place where we do camelCasing. This goes for the other abstract classes as well.

include/engine/input/i_input_provider.h
```diff
@@ -19,6 +21,9 @@ class IInputProvider {
 public:
     virtual ~IInputProvider() = default;
 
+    void set_input(std::unique_ptr<IInput> input);
```
ThomasRuijters, line 24: if you have a function called input() for getting, just name it input for setting. Its property syntax and more common in modern C++. Its also what we do throughout the rest of the engine. Only use get and set if you are not managing a property but fetching data or describing a state change that is more explicit than a value change.

ThomasRuijters (changes requested):
Decent overall. It could go for some extra comments. Change the get set thing I marked and you are good to go. If you don't want to, at least use get_prop instead of just prop as the function name.

LarsDevans:
@ThomasRuijters this reply applies to both inline comments.

It's really confusing. On the right-hand side people say the naming and syntax is correctly done, while people on the left-hand side say it isn't. I suggest we leave this up for discussion, but not here. If the linter accepts it, it's conform syntax. If there isn't something else, I'd like to merge this PR.

LarsDevans:
@ThomasRuijters I will add comments before merging

ThomasRuijters:
> @ThomasRuijters this reply applies to both inline comments.
> 
> It's really confusing. On the right-hand side people say the naming and syntax is correctly done, while people on the left-hand side say it isn't. I suggest we leave this up for discussion, but not here. If the linter accepts it, it's conform syntax. If there isn't something else, I'd like to merge this PR.

You use input() and set_input(...). Use get_ set_ or input() and input(...). Consistency matters here

LarsDevans:
@ThomasRuijters the `func` and `set_func` is, for the parental branch at least, the consistency up to this point. Getters don't get any prefix, setters do. Let's wrap this up tomorrow during the check in

LarsDevans:
> @ThomasRuijters the `func` and `set_func` is, for the parental branch at least, the consistency up to this point. Getters don't get any prefix, setters do. Let's wrap this up tomorrow during the check in

Whatever the result may be, it will be done in another PR

PR #72, opened 2025-11-13
feat(networking): skeleton

### LINKED ISSUE #52: Networking service skeleton
### Summary
Create an implementation of the `MultiplayerService` system, responsible for managing connections with peers and handle incoming and outgoing traffic to peers. The service is usable on its own, but needs a component to actually have an impact.

### Goals
- Define the `MultiplayerService` class and register it with the Engine Services Registry so it can be globally accessed.
- Define the `Host` class which acts as the host of the connection.
- Define the `Client` class which connects to a host.
- Define the `Router` class which routes incoming network messages to the corresponding handlers.
- Define the `NetworkMessage` class which contains all possible (expandable) types of network messages.
- Define the `NetworkState` class which represents the current state of a networking object.

### Deliverables
- Header and implementation files for:
    - `MultiplayerService`
    - `Host`
    - `Client`
    - `Router`
    - `NetworkMessage`
    - `NetworkState`
- Registration of the multiplayer service within the engine’s service registry.
- Documentation/comments outlining how the given methods can be utilized.

SjulGr:
Add a rough skeleton of the networking module. Due to the sheer size and complexity of the networking module, the skeleton is a rough sketch of what the final product will look like. A lot of code requires to be written in order to find out what the best solution is for several problems, which cannot be thought out from the skeleton alone.

Any inconsistencies or missing methods will be added alongside the implementation of the networking module.

Closes #52

### DIFF
--- a/include/engine/network/client.h
+++ b/include/engine/network/client.h
@@ -1,0 +1,34 @@
+#pragma once
+
+#include <memory>
+#include <enet/enet.h>
+
+#include <engine/network/router.h>
+
+class Client {
+public:
+    Client(std::shared_ptr<Router> router);
+
+    /* @brief polling network for new messages. */
+    void poll() noexcept;
+
+    /* @brief sending message to host. */
+    void send(Message& message);
+
+    /* @brief Connect to given address. */
+    void connect(std::string address);
+
+    /* @brief Disconnect from current peer. */
+    void disconnect();
+
+    /* @brief Called on connection, setting local uuid and forwarding to custom handlers. */
+    void on_connect() noexcept;
+
+    /* @brief Called on disconnection, cleaning left-over data and forwarding to custom handlers. */
+    void on_disconnect() noexcept;
+
+private:
+    std::shared_ptr<Router> router_{nullptr};
+    std::unique_ptr<ENetPeer> server_peer_{nullptr};
+    std::unique_ptr<ENetHost> client_{nullptr};
+};
--- a/include/engine/network/multiplayer_service.h
+++ b/include/engine/network/multiplayer_service.h
@@ -1,0 +1,45 @@
+#pragma once
+
+#include <memory>
+
+#include <engine/core/iEngineService.h>
+#include <engine/network/client.h>
+#include <engine/network/host.h>
+#include <engine/network/router.h>
+
+/**
+ * @brief Service containing all functions to properly manage a peer-to-peer connection.
+ *
+ * The service contains delegator methods for all useful functionalities.
+ */
+class MultiplayerService : public IEngineService {
+public:
+    /* @brief Register handler function to message type. */
+    void register_handler(const MessageType type, const std::function<void()>& handler);
+
+    /* @brief Unregister handler of given message type */
+    void unregister_handler(const MessageType type);
+
+    /* @brief Register self as host. */
+    void set_host();
+
+    /* @brief Register self as client. */
+    void set_client();
+
+    /* @brief Poll network for new messages. */
+    void poll();
+
+    /* @brief If client, send to host; If host, broadcast to all clients. */
+    void send(Message& message);
+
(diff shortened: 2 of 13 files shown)


include/engine/network/client.h
```diff
@@ -0,0 +1,34 @@
...
+    void disconnect();
+
+    /* @brief Called on connection, setting local uuid and forwarding to custom handlers. */
+    void on_connect() noexcept;
+
+    /* @brief Called on disconnection, cleaning left-over data and forwarding to custom handlers. */
+    void on_disconnect() noexcept;
+
+private:
+    std::shared_ptr<Router> router_{nullptr};
```
BotSquirrel, line 31: This could perhaps be a `reference_wrapper` to a `unique_ptr` so that its created once and shared, but I'm not sure if it has any benefits. The router would only be used in `poll` and currently has no reason to be a single instance.

include/engine/network/multiplayer_service.h
```diff
@@ -0,0 +1,45 @@
...
+    void register_handler(const MessageType type, const std::function<void()>& handler);
+
+    /* @brief Unregister handler of given message type */
+    void unregister_handler(const MessageType type);
+
+    /* @brief Register self as host. */
+    void set_host();
+
+    /* @brief Register self as client. */
+    void set_client();
```
BotSquirrel: A comment that only one of them can be active would be nice?

include/engine/network/network_state.h
```diff
@@ -0,0 +1,32 @@
...
+
+/**
+ * @brief A class containing the current state of all networking objects.
+ *
+ * The class manages the current network state of all networking objects and syncs them
+ * with the other networking peers.
+ */
+class NetworkState {
+public:
+    void set(const std::string& key, const std::any& value) noexcept;
```
BotSquirrel, line 15: Depending on how we want to serialize this we might need to narrow this `value` down. From my own experience I found that primitive types such as string, floats, bools ,etc. can be send over the net without issue. Complex types such as structs not so much.

BotSquirrel (approved):
I have some small things that could be changed but are not neccessary as the implementation of this is difficult to image at times. Some things we'll simply see/change once the usage is a bit clearer

include/engine/network/network_state.h
ThomasRuijters: You can limit this with a concept or a bit of TMP magic for auto generated types. There are types that you can use that specify that anything passed has to be a primitive, or that it has to be of a specific set of types. You can use https://en.cppreference.com/w/cpp/utility/variant.html as an example for fine grained control over an exhaustive set of types.

include/engine/util/uuid.h
```diff
@@ -0,0 +1,17 @@
...
+#include <random>
+#include <sstream>
+#include <iomanip>
+#include <array>
+#include <iostream>
+
+namespace uuid {
+    static std::random_device rd;
+    static std::mt19937 gen(rd());
+    static std::uniform_int_distribution<> dis(0, 255);
```
ThomasRuijters, line 12: Please mind that if we are going to do this... just create a random helper unless we are not doing standard randomization. I am not saying that you should, if you are doing something specific here keep it here.

ThomasRuijters (approved):
Gets my thumps up. Do mind that you are using a shared_ptr at the moment. If lifetime is not determined by multiple objects (you lifetime is determined by more than one other object) please use a unique_ptr or a reference wrapper if needed. Even a normal pointer if you have to. But smart_ptrs like unique, shared, they represent lifetime and resource ownage. Keep that in mind. weak_ptr is not a weak_reference from Java. They say something else.

SjulGr (commented):
The PR will be merged. All comments I haven't responded to will be looked back on in the future when implementing the associated parts of the code.

I also have questions in regard to some comments which I will talk about after the next stand-up.

PR #74, opened 2025-11-13
@feat/#23 rendering

### LINKED ISSUE #23: Rendering
Focusses on implementing rendering into the engine.

Subtasks are:
1. Creating SDL wrappers for textures and the renderer
2. Creating the rendering manager
3. Expanding the resource manager to support texture pooling
4. Adding the sprite component

Deliverables:

A working renderer with Texture object wrapper and sprite component, that allows the user to create sprites from buffered textures from the texture pool.

ThomasRuijters:
Adds the very basis for rendering. Includes:

1. Being able to render textures
2. Being able to register spritesheets and individual textures
3. A resource manager for storing textures efficiently
4. As an addon: The ability to manipulate the base game window

Still on the TODO are:

1. Layering is not yet supported in the renderer. This should just be a matter of simple ordering but was out of scope for the basis of rendering. I expect nothing to change in the rendering interface because of this, so it is something that we can do after. For now, this basis should be sufficient enough so that we can use it for testing and basic development.
2. The asset manager does not auto-load resources on creating the AssetService. This can either be done from the service itself or via an engine init. This is trivial and still undecided.


src/core/rendering/renderer.cpp
```diff
@@ -0,0 +1,120 @@
...
+                           static_cast<Uint8>(color.r),
+                           static_cast<Uint8>(color.g),
+                           static_cast<Uint8>(color.b));
+    SDL_SetTextureAlphaMod(texture, static_cast<Uint8>(color.a));
+}
+
+void Renderer::render(const std::vector<std::reference_wrapper<GameObject>>& objects) const{
+    SDL_RenderClear(sdl_renderer_.get());
+
+    for (auto game_obj_wrapper : objects) {
```
BotSquirrel, line 87: Now that I look at this again I realize this might profit from a visitor pattern with an interface for drawables so we can do something like `component.visit(*this)`. We would need a small refactor and a specifc SDL implementation for sprites. Basically this but moved somewhere else. 

The reason I am saying this is because right now we only focussed on textures as in objects, but what about UI? Anyway this branche is really needed at this point so the person who will be handling UI might need to change this a bit

src/core/rendering/renderer.cpp
```diff
@@ -0,0 +1,120 @@
...
+
+void set_color(const Color& color, SDL_Texture* texture) {
+    SDL_SetTextureColorMod(texture,
+                           static_cast<Uint8>(color.r),
+                           static_cast<Uint8>(color.g),
+                           static_cast<Uint8>(color.b));
+    SDL_SetTextureAlphaMod(texture, static_cast<Uint8>(color.a));
+}
+
+void Renderer::render(const std::vector<std::reference_wrapper<GameObject>>& objects) const{
```
BotSquirrel, line 84: Minor reminder for later:
This does not take the `is_active` and `is_active_in_world_` from gameobjects and components into consideration. The argument can be made that you'd filter them out before you pass them, but that logic can easily be written here and should be here.

I don't neccessarily expect that its patched now but good to bring attention to.

BotSquirrel (approved):
I do see some minor things that would present themselves when we start implementing/testing. We do need this branche so I'm fine with it for now so long as its eventually fixed

src/core/rendering/renderer.cpp
ThomasRuijters: Could do for sure. An SDL visitor or strategy could work here. The problem is that this would mean that the renderer is no longer in full control over what happens. This isn't bad by default and may actually be valid here, but the inversion of the dependency and the responsibility may actually cause friction in how we reason about working with the thing. If I tell a renderer to render something I am not telling it how to do it. I am giving it something and telling it to draw it. I don't hand it the brush. But making this more flexible could be a way to get SDL out of there, which is always nice.

src/core/rendering/renderer.cpp
ThomasRuijters: It also does not care for layering yet. It is on the patch list yes.

SjulGr (approved):

Issue #45, opened 2025-11-05
Input system


LarsDevans:
Parental task for all input related tasks

PR #78, opened 2025-11-15
feat(input): add manager + concrete system

LarsDevans:
See #45 for further details.

This PR is extra validation of the input system, 90% of code is already verified in the following PR's: #29 , #41 , #47 , #49 , #50 , #51 , #60 , #71 (lot more than I thought actually)


BotSquirrel (approved):
Nice seperation in the end. The only thing I could still think of is do we link it in the engine or something like that? I don't neccessarily need an explanation of how, but just to make sure you thought about the final usage. I assume we'll plug SDL in the main loop for checking input and then call the desired functions in behavior?

LarsDevans:
@BotSquirrel some engine service initialization method to register the input manager with the desired system. After, the engine loop calls the input's update method each frame. After that, each behavior can do, for example,
```cpp
if (input.get_key_down(KeyCode::space)) { /* Jump! */ }
```
It's that simple, or am I mistaken?

ThomasRuijters (dismissed):
Please split up the PR in smaller bits. You give me 21 files to look at here and quite a few code changes.

CMakeLists.txt
```diff
@@ -3,6 +3,7 @@ project(CapycoreEngine)
 set(CMAKE_CXX_STANDARD 23)
 set(CMAKE_CXX_STANDARD_REQUIRED ON)
 set(CMAKE_CXX_EXTENSIONS OFF)
+set(CMAKE_EXPORT_COMPILE_COMMANDS ON)
```
ThomasRuijters, line 6: Isn't this also being done inside of another PR?

include/engine/input/i_input.h
ThomasRuijters, line 1: Isn't the naming convention that I is part of the primary name? We named it after C# interfaces (do not ask me why). Its like saying IAbstraction. Its not as much camel case as it is convention. If we do it this way we will have to rename the IEngineService file as well.

include/engine/input/mouse_state.h
```diff
@@ -0,0 +1,71 @@
...
+    right   ///< Scroll wheel moved rightward.
+};
+
+/**
+ * @brief Represents per-frame mouse button transition state.
+ *
+ * Tracks both current and previous button states to detect
+ * presses and releases deterministically.
+ */
+struct MouseButtonState {
```
ThomasRuijters, line 28: Why use a struct here? There is a function on it, why not a class instead?

include/engine/input/mouse_state.h
```diff
@@ -0,0 +1,71 @@
...
+    void reset();
+};
+
+/**
+ * @brief Represents per-frame scroll state of the mouse wheel.
+ *
+ * Captures accumulated position, per-frame delta, and last known direction.
+ * Designed for deterministic querying without backend-specific dependencies.
+ */
+struct MouseWheelState {
```
ThomasRuijters, line 41: Why use a struct here? There is a function on it, why not a class instead?

include/engine/input/mouse_state.h
```diff
@@ -0,0 +1,71 @@
...
+    void reset();
+};
+
+/**
+ * @brief Aggregates all mouse-related state information.
+ *
+ * Encapsulates position, button states, and wheel movement into a single structure.
+ * Intended for per-frame input tracking and polling through IInputProvider.
+ */
+struct MouseState {
```
ThomasRuijters, line 59: Why use a struct here? There is a function on it, why not a class instead?

src/input/strategy/sdl_input_strategy.cpp
ThomasRuijters, line 1: This must have hurt...

ThomasRuijters (approved):
I would like some clarifications on the structs. But other than that, looks good to me.

CMakeLists.txt
LarsDevans: Yes

src/input/strategy/sdl_input_strategy.cpp
LarsDevans: It's generated :)

LarsDevans:
@ThomasRuijters answer to all questions relating to the choice of a struct: it are data holding types, with a simple `reset()` method to reset their state. It's not intended for any further logic.

include/engine/input/i_input.h
LarsDevans: I don't mind what prefix we choose. But when there's an argument, I believe using `i_` is easier to index interface files, because of the `i_` prefix. When it's just `i`, you have to search for `i[A-Za-z_]*` (assuming the file finder accepts regex). Additionally, snake_case is what we decided on

LarsDevans: Now that I say this, I don't believe we agreed on the casing of file names, just classes, methods/functions and variables

ThomasRuijters:
> @ThomasRuijters answer to all questions relating to the choice of a struct: it are data holding types, with a simple `reset()` method to reset their state. It's not intended for any further logic.

Yes but data holding types are not per definition structs. A struct describes structure, meaning it contains no functionality. A struct is a struct when it has, it does not do. Its like saying record values are structs because they represent data. That is simply not true. 

Use structs when exposing a C interface or when explicitly working with pure raw data. It's not idiomatic to use them the way you are. Structs and classes in C++ have so little difference between them it barely matters which one you pick. Consistency and idiom win the race here. It has functionality, its not pure data, make it a class instead or make it immutable.

PR #79, opened 2025-11-15
fix (component): no gameobject constructor and deltatime

BotSquirrel:
Fix sprite component broken constructor and missing deltatime


ThomasRuijters (approved):
I will approve of this. But why in the gods names are we removing the parent parameter?

LarsDevans:
@ThomasRuijters it's a hotfix. A dangeroes hotfix...

image

LarsDevans (approved):

PR #80, opened 2025-11-15
build(cmake): export compile commands to enchance lserver

LarsDevans:
The _clangd_ language server (the one I use) does exclusively work with this configuration is enabled

For more information read: https://cmake.org/cmake/help/latest/variable/CMAKE_EXPORT_COMPILE_COMMANDS.html


ThomasRuijters:
What exactly will this enable us to do? Is this just for your machine? I can't tell even by looking at the link

LarsDevans:
It's the thing that says to the language server "Hey, here's the definition you're looking for"
Basically when it's not there, no definitions are found, and it are all errors. See below:
image

SjulGr (approved):

ThomasRuijters (approved):

Issue #81, opened 2025-11-15
Vector3 extended arithmetic


LarsDevans:
Currently, `Vector3` supports basic addition and subtraction with `operator+`, `operator-`, `operator+=`, and `operator-=`. We need to expand this to fully support all arithmetic operations efficiently, including:

- `operator*` and `operator/` with both scalars and other `Vector3` objects
- Compound assignment operators: `*=`, `/=`

**Tasks:**

- [x] Add `operator*`, `operator/` for `Vector3` - `Vector3`
- [x] Add `operator*`, `operator/` for `Vector3` - `float`
- [x] Add compound assignment operators `*=`, `/=` for both cases
- [x] Remove inefficient rvalue reference overloads (`Vector3&&`) where unnecessary
- [x] Validate all operators with Catch2 unit tests:

  - Test basic arithmetic (`+`, `-`, `*`, `/`)
  - Test compound assignment (`+=`, `-=`, `*=`, `/=`)
  - Test scalar multiplication/division
  - Test combinations of `Vector3` operations

PR #82, opened 2025-11-15
feat(vector3): extend the arithmetic operations

### LINKED ISSUE #11: #10-Animate objects via calculations
Als een gamedeveloper/programmeur
Wil ik een gameobject eenvoudig vanuit code kunnen laten schalen, draaien (roteren) en verplaatsen (transleren) middels berekeningen.
Zodat ik voor dat soort eenvoudige animaties geen spritesheet hoef te gebruiken.

LarsDevans:
See #81 for the details

### DIFF
--- a/include/engine/public/util/vector3.h
+++ b/include/engine/public/util/vector3.h
@@ -1,15 +1,27 @@
 #pragma once
+
 class Vector3 {
 public:
-    Vector3();
-    Vector3(float x, float y, float z);
     float x;
     float y;
     float z;
 
-    Vector3 operator+(const Vector3& right) const noexcept;
-    Vector3 operator-(const Vector3& right) const noexcept;
+    Vector3();
+    Vector3(float x, float y, float z);
 
-    Vector3& operator+=(const Vector3&& right) noexcept;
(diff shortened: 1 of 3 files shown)


ThomasRuijters (commented):
Why exactly is this needed? I see the feature branch sure, but where are we using multiplication and division. you do not need it for animation at the moment. Velocity and applied force is just a vector that is summed onto the current position of an entity. I do not see why we need multiplication and division.

ThomasRuijters (changes requested):
At the moment I do not see the need for the change. I would like a proper description and proposal for why this is needed. I feel like it does nothing at the moment. Vec3 is a constant position or a change in position, not a factor applied to one. The only reason you would want multiplication is if you skip physics frames and say "Lets just take the last 5 updates, skip them, and make the thing move one giant chunk". There is no real need for it yet.

ThomasRuijters (approved):
The code itself looks fine, but I do not see why we need it at the moment. Nice addition, its clear now that you are not speaking about animation but just nice additions to the vector3 class. I'll approve of this.

SjulGr (approved):

PR #86, opened 2025-11-16
feat(game object): add uuid

ThomasRuijters:
Adds UUID to gameobject, exposed via ID() property

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,5 +1,6 @@
 cmake_minimum_required(VERSION 3.25)
 project(CapycoreEngine)
+
 set(CMAKE_CXX_STANDARD 23)
 set(CMAKE_CXX_STANDARD_REQUIRED ON)
 set(CMAKE_CXX_EXTENSIONS OFF)
@@ -107,20 +108,6 @@ set(IMGUI_SOURCES
 )
 
 # =========================================================
-# Boost (header-only UUID + required sub-libs)
-# =========================================================
-set(BOOST_INCLUDE_LIBRARIES uuid random)   # Only what we need
-set(BOOST_ENABLE_CMAKE ON CACHE BOOL "Enable Boost CMake build")
-
-FetchContent_Declare(
-        boost
-        GIT_REPOSITORY https://github.com/boostorg/boost.git
-        GIT_TAG boost-1.86.0
-)
-
-FetchContent_MakeAvailable(boost)
-
-# =========================================================
 # Testing
 # =========================================================
 Set("ENABLE_TESTS" ON CACHE BOOL "Enable building tests")
@@ -151,8 +138,6 @@ target_include_directories(engine
 
 target_link_libraries(engine
     PUBLIC
-        Boost::uuid
-        Boost::random
         SDL3::SDL3
         SDL3_image::SDL3_image
         SDL3_ttf::SDL3_ttf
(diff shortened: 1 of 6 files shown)


LarsDevans (approved):

BotSquirrel (approved):
If we all agree that an extra lib is fine then sure but personally I don't see a reason to fetch `Boost` and just include those two. We do still download and store the entire lib. The UUID does not have to be cryptographically secure, if we can ensure enough randomness then its fine for our usecase.

e.g
```cpp
namespace uuid {

    inline thread_local std::mt19937 gen(std::random_device{}()); // should cover us fine
    inline thread_local std::uniform_int_distribution<uint8_t> dis(0, 255);

    inline std::string generate_uuid_v4() {
        std::array<uint8_t, 16> bytes;

        for (auto& b : bytes) {
            b = dis(gen);
        }

        bytes[6] = (bytes[6] & 0x0F) | 0x40;
        bytes[8] = (bytes[8] & 0x3F) | 0x80;

        char buffer[37]; // 36 chars and a null
        snprintf(
            buffer, sizeof(buffer),
            "%02x%02x%02x%02x-%02x%02x-%02x%02x-%02x%02x-%02x%02x%02x%02x%02x%02x",
            bytes[0], bytes[1], bytes[2], bytes[3],
            bytes[4], bytes[5],
            bytes[6], bytes[7],
            bytes[8], bytes[9],
            bytes[10], bytes[11], bytes[12], bytes[13], bytes[14], bytes[15]
        );

        return std::string(buffer);
    }
}
```

ThomasRuijters:
> ```c++
> ine std::string generate_uuid_v4() {
>         std::array<uint8_t, 16> bytes;
> 
>         for (auto& b : bytes) {
>             b = dis(gen);
>         }
> 
>         bytes[6] = (bytes[6] & 0x0F) | 0x40;
>         bytes[8] = (bytes[8] & 0x3F) | 0x80;
> 
>         char buffer[37]; // 36 chars and a null
>         snprintf(
>             buffer, sizeof(buffer),
>             "%02x%02x%02x%02x-%02x%02x-%02x%02x-%02x%02x-%02x%02x%02x%02x%02x%02x",
>             bytes[0], bytes[1], bytes[2], bytes[3],
>             bytes[4], bytes[5],
>             bytes[6], bytes[7],
>             bytes[8], bytes[9],
>             bytes[10], bytes[11], bytes[12], bytes[13], bytes[14], bytes[15]
>         );
> 
>         return std::string(buffer);
>     }
> }
> ```

Could do, but we would have to test this. I will hit you up IRL for this, we can get this to work within 30 min easy.

CMakeLists.txt
```diff
@@ -106,6 +106,20 @@ set(IMGUI_SOURCES
...
+set(BOOST_ENABLE_CMAKE ON CACHE BOOL "Enable Boost CMake build")
+
+FetchContent_Declare(
+        boost
+        GIT_REPOSITORY https://github.com/boostorg/boost.git
+        GIT_TAG boost-1.86.0
+)
+
+FetchContent_MakeAvailable(boost)
+
```
BotSquirrel: Nice, I see its hooked in gameobject so this can be gone too?

BotSquirrel (changes requested):

BotSquirrel (approved):

SjulGr (approved):

Issue #69, opened 2025-11-12
Physics component(s)


BotSquirrel:
The implementation of the public physics components utilizing the service. The following components will be available:

* `Rigidbody2D`
* `Collider2D`
* `BoxCollider2D`
* `CircleCollider2D`

PR #87, opened 2025-11-17
feat (physics): add colliders and rigidbody component

BotSquirrel:
See #69 🤡 for more info. TLDR; It adds a rigidbody2d, collider2d, box collider2d and circle collider2d using the physicsservice. I had to extent the service a lot more then i thought so sorry about the size... However, most of this is more extending of the service due to needing a lot more than i figured at first


ThomasRuijters:
You know I will review this, but you it will hurt me. 36 files, this pain will compensate for the hour difference 😡

include/engine/physics/world/body/body_2d.h
ThomasRuijters, line 1: Why exactly do it this way instead of using regular method class syntax?

include/engine/public/components/colliders/box_collider_2d.h
```diff
@@ -0,0 +1,41 @@
...
+ * enabling rectangular collision detection and response within the physics engine.
+ */
+class BoxCollider2D : public Collider2D
+{
+public:
+    BoxCollider2D(
+        float friction, 
+        float bounciness,
+        float width = 1.0f,
+        float height = 1.0f
```
ThomasRuijters: Magic numbers

include/engine/public/components/colliders/box_collider_2d.h
```diff
@@ -0,0 +1,41 @@
...
+    [[nodiscard]]
+    float height() const noexcept;
+    BoxCollider2D& height(float value) noexcept;
+
+    BoxCollider2D& friction(float value) noexcept override;
+    BoxCollider2D& bounciness(float value) noexcept override;
+
+private:
+    float width_ {1.0f};
+    float height_ {1.0f};
```
ThomasRuijters: Magic numbers

include/engine/public/components/rigidbody_2d.h
```diff
@@ -0,0 +1,78 @@
...
+ * This component allows a GameObject to interact with the physics engine,
+ * enabling features such as gravity, collisions, and movement based on physical forces.
+ */
+class Rigidbody2D : public Component
+{
+public:
+    Rigidbody2D(BodyType2D::Type type = BodyType2D::Dynamic, 
+                float mass = 1.0f,
+                bool use_gravity = true,
+                float gravity_scale = 1.0f);
```
ThomasRuijters: Magic values? Or is this fine here?

src/public/components/colliders/box_collider_2d.cpp
```diff
@@ -0,0 +1,104 @@
...
+#include <engine/public/gameObject.h>
+
+BoxCollider2D::BoxCollider2D(
+    float friction, 
+    float bounciness,
+    float width,
+    float height
+) : Collider2D(friction, bounciness), width_(width), height_(height) 
+{
+    add_on_attach([this](Component& comp) {
```
ThomasRuijters: Could you split this lambda into a separate function? I get that you only use it here but it looks a bit bloated to me.

ThomasRuijters (approved):
Some minor comments, but looks good to me. I can not really test this on my end so I have to assume that this has been checked. 32 files is to much to be specific about or else I would be finishing this review at the end of the week.

include/engine/physics/world/body/body_2d.h
BotSquirrel: `body_2d` was never imagened as a class but as a simple data passage. Box2D made both shape and body ID's simple types not meant for copying/pointers. I made it as a simple passable data type. Later I figured out I need to pass a lot through simple box2d calls. I figured Body2D was a nice way to group everything together.

SjulGr (approved):

PR #96, opened 2025-11-17
feat(scene): add basic + scene manager

ThomasRuijters:
The basis for the scene and scene manager has been written. It turns out that I made a mistake and build both features on the same branch. Turns out it wasn't a lot. A couple of things I found during the making of the branch:

We have:
```cpp
void Scene::on_awake() {

}

void Scene::destroy() {

}
```

It does nothing yet, not until we decide if they are hooks or basic over-writable methods. I vote for making them hooks. Just give it an on_pause, on_awake, on_destroy hook. Though I do not see why we would want an on_destroy hook.

On a second note, we can not move scenes. Ones you make them, you make them. If you want your own custom scene you will have to inherit from the thing. You can instantiate one inline, for sure. But if you do so, when you move it you move the thing with its vector of unique_ptr instances. C++ does not agree. It could cause issues, so perhaps we need to define a move constructor if this is something we want to be able to do. I personally am a fan of using constructor functions or just using inheritance in this case (less favorable I suppose, inheritance usually evil). It really depends on how the user wants to store scenes or swap them.


CMakeLists.txt
```diff
@@ -4,7 +4,7 @@ project(CapycoreEngine)
 set(CMAKE_CXX_STANDARD 23)
 set(CMAKE_CXX_STANDARD_REQUIRED ON)
 set(CMAKE_CXX_EXTENSIONS OFF)
-set(CMAKE_EXPORT_COMPILE_COMMANDS ON)
+# set(CMAKE_EXPORT_COMPILE_COMMANDS ON)
```
BotSquirrel: Best to leave this on for others as it shouldn't create a problem?

include/engine/public/scene.h
```diff
@@ -1,5 +1,41 @@
...
+    void stop();
+
+    Scene& time_scale(float modifier);
+    [[nodiscard]] float time_scale() const;
+
+    [[nodiscard]] const std::string& name() const;
+    [[nodiscard]] bool is_running() const;
+
+    [[nodiscard]]
+    std::reference_wrapper<GameObject> get_game_object(const std::string& name) const;
```
BotSquirrel: seeing as we now use ID's in gameobjects, is that not a better search param? (apparently you do that later but named it wrong here)

src/public/gameObject.cpp
```diff
@@ -74,6 +81,11 @@ bool GameObject::is_active() const noexcept {
 
 GameObject& GameObject::parent(GameObject& parent) {
     parent_ = parent;
+    if(std::find_if(parent.children().begin(), parent.children().end(),[&](auto& ref) {
```
BotSquirrel: A small thing but perhaps better to move to a var and compare?

src/public/scene.cpp
```diff
@@ -1 +1,133 @@
...
+
+}
+
+void Scene::destroy() {
+
+}
+
+void Scene::game_loop() {
+    float accumulator = 0.0f;
+    const float fixed_step = 1.0f / 60.0f; // ~60 fps
```
BotSquirrel: magic numbers?

src/public/scene.cpp
```diff
@@ -1 +1,133 @@
...
+
+    Uint64 last = SDL_GetPerformanceCounter();
+    float freq = static_cast<float>(SDL_GetPerformanceFrequency());
+
+    auto& rendering_service = Engine::instance()
+        .services
+        ->get_service<RenderingService>()
+        .get();
+
+    while (true) {
```
BotSquirrel: `while(is_running_)`? How do we exit this otherwise?

src/public/scene_service.cpp
```diff
@@ -0,0 +1,17 @@
...
+
+const std::string& default_scene_name = "default";
+
+SceneService::SceneService()
+    : scene_(std::make_unique<Scene>(default_scene_name)) {}
+SceneService::SceneService(std::unique_ptr<Scene> initial_scene)
+    : scene_(std::move(initial_scene)) {}
+
+SceneService::~SceneService() = default;
+SceneService& SceneService::load_scene(std::unique_ptr<Scene> scene) {
```
BotSquirrel: Shouldn't we tag it as `running`? I am fine if we don't and say that we load them default as not running but then we'd need a switch or start method

include/engine/public/scene.h
```diff
@@ -1,5 +1,41 @@
...
+    std::vector<std::unique_ptr<GameObject>> game_objects_;
+
+    void game_loop();
+
+public:
+    Scene(const std::string& name);
+    virtual ~Scene();
+
+    void on_awake();
+    void destroy();
```
BotSquirrel: I see your point and hit the same topic with components `on_attach`/`on_detach`. In my opinion its better to keep 2 vectors with functions (`std::function<void(Scene&)`) that are triggered in these calls. If we make an `add_on_awake(...)` we can dynamically add lamdba's from our behaviorscript. Might be a bit overkill but nice to have that possibility as game dev and not too much work?

BotSquirrel (changes requested):
I would like some of the things changed if possible. Also as I ran into memory leaks upon testing the physics components I ran my own little test.

When making a scene like below and not adding it to the service, it will leak. Only if we add it does it not leak. Its a minor thing thats optional as we problaby always put it in the service but might be worth to quickly check

```cpp
Engine& engine = Engine::instance();

std::unique_ptr<Scene> scene = std::make_unique<Scene>("Main Scene");
GameObject& player = scene->add_game_object("Player");
GameObject& enemy = scene->add_game_object("Enemy");

// auto& scene_service = engine.services->get_service<SceneService>().get();
// scene_service.load_scene(std::move(scene));

engine.quit();
```

Lastly on the note of copying/moving. In my opinion scenes should be managed within the service, not by the user. You simple registrate and ask to switch. We'll do the rest so we are in control of that. I think not copying/moving is fine

CMakeLists.txt
ThomasRuijters: Agreed, I put it into comment because I was trying to figure out why stuff did not want to build. Just forgot to take it back out of purgatory. Good find.

include/engine/public/scene_service.h
```diff
@@ -0,0 +1,20 @@
+#pragma once
+
+#include <memory>
+#include <engine/public/scene.h>
+#include <engine/core/iEngineService.h>
+
+class SceneService final : public IEngineService{
+private:
+    std::unique_ptr<Scene> scene_;
```
BotSquirrel: So why do we store 1 scene? I pictured this as a service for containing and switching. If we just keep 1 scene then there isn't really a reason for this to exist

include/engine/public/scene.h
ThomasRuijters: I used the name as an ID everywhere but forgot to change the function param name. Will get to it.

src/public/gameObject.cpp
ThomasRuijters: Could do, but we do not need the extra allocation. Its just a ptr so I won't complain about 8 bytes tbf. Will do.

src/public/scene.cpp
ThomasRuijters: Touche

src/public/scene.cpp
ThomasRuijters: Do not know how that made it in. Should have been removed last commit, I will see what is up with that

include/engine/public/scene_service.h
ThomasRuijters: That is what I called for earlier in the the design phase as well. There are a couple of problems with this. Imo there are three main ways to create scenes. You create one on the spot as you set the current one, you prep an instance before hand and feed it to the scene service, and you just inherit and make one inline as you set it via the service. A couple of things to note here:

Inheritance is something I usually want to avoid. In this case however it does fit the theme of what unity tends to do. You do not really store instances of scenes in the user code. You spawn one on the spot and that is it. Inheritance solves for the lack of ability to alter the scene before you would load it into the manager. I see a couple of downsides here but bottom line is that the (in my opinion) better alternative is using composition and factory functions instead. That way you can create a scene, store it, and later load it back in. 

The problem with my preferred approach at the moment is that the moving of a scene is kind of prohibited. It holds a vector of unique_ptr and those things to not like being moved much, it results in big kaboom your code will hate you. You would always have to wrap scenes inside of unique_ptr instances and you would have to transition ownership of the thing between the manager and the code storing the instance. Now a way to (again) solve for this is to just use naming instead and letting the manager deal with the owning of multiple scenes. I would like to know your thoughts on this.

src/public/scene.cpp
```diff
@@ -1 +1,146 @@
...
+
+    float accumulator = accumulator_default_value;
+
+    Uint64 last = SDL_GetPerformanceCounter();
+    float freq = static_cast<float>(SDL_GetPerformanceFrequency());
+
+    auto& rendering_service = Engine::instance()
+        .services
+        ->get_service<RenderingService>()
+        .get();
```
LarsDevans, line 46: Is it for sure this will always have a value?

src/public/scene_service.cpp
```diff
@@ -0,0 +1,93 @@
...
+    if(!current_scene_opt.has_value()) {
+        it->second->run();
+        return *this;
+    }
+
+    Scene& current_scene = current_scene_opt.value().get();
+
+    if(current_scene.name() == name) {
+        return *this; // Scene is already loaded
+    }
```
LarsDevans, line 46: We could use this block here to reload the scene. Do we want what? Is there a scenario where to game developer wants this?

tests/engine/scenes/scene.cpp
```diff
@@ -0,0 +1,123 @@
...
+    scene.remove_game_object(parent);
+
+    // assert
+    REQUIRE(scene.game_objects().empty());
+}
+
+// Do not make these tests. They will call the engine singleton and give you a big headache.
+//TEST_CASE("OnRunListeners_AreCalledWhenSceneRuns", "[Scene]")
+//TEST_CASE("OnStopListeners_AreCalledWhenSceneStops", "[Scene]")
+//TEST_CASE("OnDestroyListeners_AreCalledWhenSceneSceneDestroyed", "[Scene]")
```
LarsDevans, line 123: Should we not remove them either?

tests/engine/scenes/scene_service.cpp
```diff
@@ -0,0 +1,72 @@
...
+    // assert
+    REQUIRE(scene_names.size() == 2);
+    REQUIRE(scene_names.find(name1) != scene_names.end());
+    REQUIRE(scene_names.find(name2) != scene_names.end());
+}
+
+// Do not make these tests. They will work with the engine and kick off SDL windows. You will be hurt, do not.
+//TEST_CASE("LoadScene_WhenSceneExists_SetsSceneAsRunning", "[SceneService]");
+//TEST_CASE("LoadScene_WhenSceneDoesNotExists_Throws", "[SceneService]");
+//TEST_CASE("AddSceneAndLoad_WhenSceneNotExist_LoadsAndAddsScene", "[SceneService]")
```
LarsDevans, line 72: Should we not remove them either?

src/public/scene.cpp
ThomasRuijters: The rendering service will always exist. It has to.

src/public/scene_service.cpp
ThomasRuijters: The dev does not get to load and unload scenes by hand. A loaded scene is a running scene. The manager automagically sets all other scenes to stopped. The user does not get to do that by hand, only the manager.

tests/engine/scenes/scene.cpp
ThomasRuijters: I leave them there as a warning. Future me will probably be stupid and try adding them again later. A friendly warning more then anything else.

tests/engine/scenes/scene_service.cpp
ThomasRuijters: I leave them there as a warning. Future me will probably be stupid and try adding them again later. A friendly warning more then anything else.

LarsDevans (approved):

BotSquirrel (approved):
Some tests, the ones using the old gameobjects, will problably break as `Scene` is now a private constructor. We can make a fix branche for that as its on items that don't belong to this topic but feel the changes

Issue #89, opened 2025-11-17
UI components


BotSquirrel:
### Summary
Create the UI components for buttons, text and images with a Camera. This issue is a parental issues with sub issues for the specifc components and skeletons.

### Goals
* Add a general UI component interface
* Add a Text component
* Add a image component (something along the lines of health bars for instance)
* Add a Button component
* Add a camera
* A restructure for rendering?

### Deliverables
Header and .cpp files for each component with the camera

PR #98, opened 2025-11-18
 feat (UI): create UI components base skeleton

BotSquirrel:
I paused this feature #89  as I need some refactoring from someone else. However, I do think this skeleton is fine for now.


ThomasRuijters (approved):
Looks good to me

SjulGr (approved):

Issue #100, opened 2025-11-19
Behavior and behaviorscript component


BotSquirrel:
### Summary
Create the core behavior class, which can be extended to write own behavior, and the components its housed within. This is a parental issue for both items.

### Goals
* Add a public behavior class that should be extended
* Add a component for the behavior: `behaviorscript`

### Deliverables
Header and .cpp files for the behavior and its component

PR #103, opened 2025-11-19
feat (behavior): add extendable class and component

BotSquirrel:
See #100 for more info. TLDR; Add a behavior class and component so we can mutate components and objects within the scene


include/engine/public/behavior.h
```diff
@@ -0,0 +1,132 @@
...
+    virtual void on_awake() = 0;
+    virtual void on_start() = 0;
+    virtual void on_update(float dt) = 0;
+    virtual void on_destroy() = 0;
+
+    [[nodiscard]] GameObject& game_object() const;
+    [[nodiscard]] Transform& transform() const;
+
+    [[nodiscard]] bool enabled() const;
+    Behavior& enabled(bool value);
```
ThomasRuijters: In case of a boolean just use a function called enable for enable, and disable for disable. That makes the intention a lot clearer

ThomasRuijters (approved):

include/engine/public/behavior.h
```diff
@@ -0,0 +1,133 @@
...
+public:
+    Behavior();
+    virtual ~Behavior() = default;
+
+    void attach(Component& component);
+
+    virtual void on_awake() = 0;
+    virtual void on_start() = 0;
+    virtual void on_update(float dt) = 0;
+    virtual void on_destroy() = 0;
```
LarsDevans: When I was working in Unity, the awake, start and update method are optional. You don't always need them as (game)programmer. Is that also the case for us? If so, consider default empty-body implementations.

include/engine/public/gameObject.h
```diff
@@ -120,7 +120,7 @@ class GameObject {
         auto component = std::make_unique<T>(std::forward<Args>(args)...);
         T& ref = *component;
 
-        component->parent(std::ref(*this));
+        auto& comp = component->parent(std::ref(*this));
```
LarsDevans: Why is `parent(<>)` marked with _[[nodiscard]]_. It can be easily discarded here, no?

src/public/behavior.cpp
```diff
@@ -0,0 +1,82 @@
...
+    }
+}
+
+void Behavior::destroy(GameObject& game_object) 
+{
+    if (!attached_component_) {
+        throw std::runtime_error("Behavior has no associated component.");
+    }
+
+    // Get scene from parent GameObject and remove the specified game object
```
LarsDevans: If I am not mistaking, we can already achieve this by enabling the `marked_for_deletion_`. The actual deletion can be done in the scene itself

include/engine/public/behavior.h
BotSquirrel: Yeah I agree. I copied over a lot of things from monobehavior and forgot we don't always need this. Also i assume you meant `on_destroy` and not `on_update` in the list of optional methods as most the times we'd change something in update right?

include/engine/public/gameObject.h
BotSquirrel: Yes in this case it can be easily discarded, but its a linter thing. I'll supress it using a `// NOLINT`. Its not the prettiest but its either that or this and i guess the nolint is better.

include/engine/public/behavior.h
LarsDevans: Sure, not implementing the update is unusual

SjulGr (approved):

Issue #104, opened 2025-11-19
Implement save and load


LarsDevans:
### Description
The `SimpleStorage` class currently has placeholder `save()` and `load()` methods. These need proper implementation to persist and restore the in-memory key-value storage to disk.

### Tasks

- **save()**
  - Create the `saved` directory if it doesn't exist.
  - Write all key-value pairs to `saved/simple.dat`.
  - Use a simple text format where each line is `key=value`.
    - Strings should be quoted and escaped.
    - Ints and floats are written as-is.
  - Include a header warning users not to manually edit the file.
  - If writing fails, silently skip or optionally log an error.

- **load()**
  - Open `saved/simple.dat` if it exists.
  - Read line by line:
    - Ignore lines starting with `#`.
    - Parse key-value pairs exactly as saved (no trimming).
    - Strings must be unescaped.
    - Convert numeric values to int or float.
  - If a line is malformed or conversion fails, print an error with line number but continue processing remaining lines.

- **Helper functions (inside anonymous namespace)**
  - `save_path()` returns the file path `saved/simple.dat`.
  - `escape_string()` escapes `"` and `\` in strings.
  - `unescape_string()` reverses escaping.

### Notes
- Only `int`, `float`, and `string` types need to be supported.
- `load()` is private; only the engine should call it during startup.
- The file format is simple and readable; any manual modifications may break the storage and are explicitly warned against.

Issue #84, opened 2025-11-16
Simple storage skeleton


LarsDevans:
### Summary

Create the initial skeleton implementation of the `SimpleStorage` system, responsible for lightweight key-value storage within the engine. This issue focuses purely on class structure, public interface layout, and engine-level integration. No persistence backend or serialization logic should be implemented yet.

### Goals

- Provide the public API for in-memory session storage:
  - `try_get(const StorageKey&)`
  - `set_value(const StorageKey&, const StorageValue&)`
  - `delete_all()`
  - `delete_key(const StorageKey&)`
  - `save()` (stub only)
- Establish the value types supported by the storage using a `std::variant<int, float, std::string>`.
- Add a private `load()` stub to represent future persistent loading logic.

### Deliverables

- Header and minimal `.cpp` files for `SimpleStorage`, containing:
  - Full class declaration.
  - Stubbed method definitions that compile but perform no real work beyond manipulating the in-memory map.
- Inline documentation describing intended behavior

PR #105, opened 2025-11-19
feat(storage): add save and load + tests

LarsDevans:
About over 90% of code is already reviewed in another PR (#84). The thing that needs its review is `src/storage/simple_storage.cpp`.

See #104  for further details.

### DIFF
--- a/src/storage/simple_storage.cpp
+++ b/src/storage/simple_storage.cpp
@@ -1,0 +1,40 @@
+#include <engine/storage/simple_storage.h>
+
+SimpleStorage& SimpleStorage::instance()
+{
+    static SimpleStorage inst;
+    return inst;
+}
+
+std::optional<SimpleStorage::StorageValue> SimpleStorage::try_get(const StorageKey& key_name)
+{
+    if (const auto it = session_storage_.find(key_name); it != session_storage_.end()) {
+        return it->second;
+    }
+    return std::nullopt;
+}
+
+void SimpleStorage::set_value(const StorageKey& key_name, const StorageValue& value)
+{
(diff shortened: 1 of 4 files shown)


BotSquirrel (approved):
Looks nice, I assume it works universally as you create a `saves` folder using the `filesystem` library

src/storage/simple_storage.cpp
```diff
@@ -0,0 +1,171 @@
...
+        } else if (std::holds_alternative<float>(value)) {
+            out << std::get<float>(value);
+        } else if (std::holds_alternative<std::string>(value)) {
+            out << "\"" << escape_string(std::get<std::string>(value)) << "\"";
+        }
+        out << "\n";
+    }
+}
+
+void SimpleStorage::load()
```
ThomasRuijters, line 99: Could you split up the parsing and the running of the acual loading into separate functions? This thing is doing multiple things at once.

ThomasRuijters (changes requested):

LarsDevans:
@BotSquirrel I guess we will find out. But I guess because of it's integrated nature, it will work just fine

src/storage/simple_storage.cpp
LarsDevans: Do you think the effort put into that will yield it fruits? We will probably never touch this class again, so I don't believe it's worth the effort. It works fine (it has passing tests)

ThomasRuijters: Its a function. That is all. joink and paste, slap a name on it, call it. But I won't block you for it if it is to much effort.

ThomasRuijters (approved):

PR #112, opened 2025-11-21
feat (rendering): add the new rendering strategies for UI and drawables

### LINKED ISSUE #89: UI components

BotSquirrel:
This PR is a curveball from UI as I needed the renderer to support multiple methods. Instead @ThomasRuijters and myself reworked it so it utilizes strategies for drawing. That way we can decide the draw logic per component instead of function based in the renderer.

Its on the larger side but with a refactor like this its bound to happen. I left the UI components on another branche for now.


SjulGr (approved):

LarsDevans (commented):
Correct me if I'm wrong; the flow is now as follows:
`create game object` --> `add some Rendering component` --> `choose rendering strategy` --> `rendering factory creates instance` --> `inject instance to component` --> `scene draws the generic rendering strategy`

BotSquirrel:
> Correct me if I'm wrong; the flow is now as follows: `create game object` --> `add some Rendering component` --> `choose rendering strategy` --> `rendering factory creates instance` --> `inject instance to component` --> `scene draws the generic rendering strategy`

almost but the user doesn't have to do anything.
The component fetches its strategy from the factory and saves that in a member, it doesn't inject it. Its also passes itself so you'd only need to add a strategy and `if` condition

LarsDevans (approved):

BotSquirrel:
Don't delete the branche just, GUI is building on top of this.

PR #113, opened 2025-11-21
feat (networking): add connection and disconnecting

### LINKED ISSUE #53: Network connect and disconnect implementation
### Summary
Fill the skeleton for the `multiplayer service` with a working implementation to allow connecting to a peer as either a host or a client.

### Goals
- Be able to register myself as a host
- Be able to register myself as a client
- Be able to connect to a host via a local IP
- Be able to disconnect myself
- Graceful shutdown upon disconnect

Deliverables
- A working `multiplayer service` that can be used by the service container
- Functionality to connect and disconnect to peers

SjulGr:
Add the connecting and disconnecting to the networking service. The networking service contains a lot of edge-cases, so some cases might result in unknown behavior. Testing is done via a small program in the main function due to the difficulty of testing networking.

Closes #53


include/engine/network/multiplayer_service.h
```diff
@@ -1,45 +1,101 @@
...
 private:
-    std::shared_ptr<Router> router_;
-    std::unique_ptr<Client> client_;
-    std::unique_ptr<Host> host_;
+    std::shared_ptr<Router> router_{nullptr};
+    std::unique_ptr<Client> client_{nullptr};
+    std::unique_ptr<Host> host_{nullptr};
+
+    int connection_port_{1024};
+    int max_clients_{4};
```
BotSquirrel: Extract in constexpr instead of magic numbers?

include/engine/network/connection_state.h
```diff
@@ -0,0 +1,10 @@
+#pragma once
+
+enum class ConnectionState {
```
BotSquirrel: No type? I saw you do it with others like `DefaultMessage`

src/network/client.cpp
```diff
@@ -1,36 +1,167 @@
 #include <engine/network/client.h>
+#include <engine/network/host.h>
+
+#include <cstring>
+#include <stdexcept>
 
 Client::Client(std::shared_ptr<Router> router)
+    : router_(std::move(router))
 {
+    client_ = enet_host_create(nullptr, 1, 2, 0, 0);
```
BotSquirrel: magic numbers 🪄

include/engine/network/network_message.h
```diff
@@ -42,11 +45,18 @@ struct MsgDisconnect
 template <typename T>
 Message SerializeMessage(const T& data, MessageType type)
```
BotSquirrel, line 46: Not a comment that has to be applied but it would be nice if we could do something like a `concept` or simple `static_assert` to check that this can only be created for objects that inherit from one of the enums. That way we restrict someone from dumping lets say a gameobject in here (which is too complex to send)

src/network/client.cpp
```diff
@@ -1,36 +1,167 @@
...
-void Client::connect(std::string address)
+void Client::connect(const std::string& host_ip, int connection_port)
 {
+    connection_state_ = ConnectionState::CONNECTING;
+
+    ENetAddress address{};
+    enet_address_set_host(&address, host_ip.c_str());
+    address.port = connection_port;
+
+    server_peer_ = enet_host_connect(client_, &address, 2, 0);
```
BotSquirrel: magic numbers? 🪄

src/network/host.cpp
```diff
@@ -1,36 +1,227 @@
...
+    }
+}
+
+void Host::start_server()
 {
+    ENetAddress address{};
+    address.host = ENET_HOST_ANY;
+    address.port = connection_port_;
+
+    server_ = enet_host_create(&address, max_clients_, 2, 0, 0);
```
BotSquirrel: more magic 🪄

src/network/host.cpp
```diff
@@ -1,36 +1,227 @@
...
 {
+    if (!server_)
+        return;
+
+    ENetEvent event{};
+
+    while (enet_host_service(server_, &event, 0) > 0) {
+        switch (event.type) {
+
+        case ENET_EVENT_TYPE_CONNECT: {
```
BotSquirrel, line 60: Did you test this with multiple clients? I am not sure but do we want to broadcast this, as in alert everyone that someone else is connected? I try to picture the game loop in my head and if someone joins the host a body is to be  created on all clients. We could do that somewhere else or later as maybe i am overthinking this and does it not belong here.

src/network/multiplayer_service.cpp
```diff
@@ -1,41 +1,160 @@
...
 void MultiplayerService::poll()
 {
+    if (host_) {
+        host_->poll();
+    }
+    else if (client_) {
+        client_->poll();
+    }
+    else {
+        throw std::runtime_error("Cannot poll: service is neither client nor host.");
```
BotSquirrel: I am for throwing but this one might be bit dangerous. We always include this service so we practically force the user to either be the host or client. In the case of the tutorial level, it does not have to be multiplayer so do we want this?

perhaps we could introduce a bool like `is_multiplayer` that simple returns on false?

src/network/router.cpp
```diff
@@ -1,16 +1,26 @@
...
 
-void Router::route(const MessageType type, std::any& data)
+void Router::route(const Message& msg) const
 {
-
+    auto it = handlers_.find(msg.header.type);
+    if (it != handlers_.end()) {
+        it->second(msg);
+    }
+    // else: unknown message type - intentionally ignored
```
BotSquirrel, line 25: Not something you need to do but it might be interesting to discuss debug channels? It be nice later to log a few things, this is one of those I think

BotSquirrel (changes requested):
Most of it looks nice 👌, I assume you tested it as I don't have time while doing the review. The magic numbers should be fixed as its easier to read and the linter will ask you anyway, the others not so much but good to take a look at

src/network/host.cpp
SjulGr: I did test with multiple clients, yes. I had thought about broadcasting when a client connects, but haven't thought of any specific reason to. The host manages the client connections, so on a networking level, none of the clients _need_ to know of the other clients.

When a client player connects to a host player, it is up to the game developer what to send to the other clients. In the case of Capybrawla, when I connect to a lobby, the lobby needs my display name, so I will have to send a network message containing my base information. This can then be broadcasted to all clients so they know who joined and such.

src/network/router.cpp
SjulGr: Debug channels as in some kind of logger for the engine itself?

include/engine/network/connection_state.h
ThomasRuijters: Type in the name is not really needed in this case, unless we make actual instances of the classes below. The naming looks fine to me but I get the fuzz.

include/engine/network/multiplayer_service.h
ThomasRuijters: Magic numbers, please use constexpr int_16 here.

include/engine/network/network_message.h
ThomasRuijters: I do not know if that is possible. You can write a concept that checks if a certain member is present. Do you mean that you want to enforce that T is of a type that is serializable over the network?

include/engine/network/multiplayer_service.h
```diff
@@ -1,45 +1,101 @@
...
+    [[nodiscard]] int get_client_amount() const noexcept;
+
+    void set_connection_port(int port) noexcept;
+    [[nodiscard]] int get_connection_port() const noexcept;
+
 private:
-    std::shared_ptr<Router> router_;
-    std::unique_ptr<Client> client_;
-    std::unique_ptr<Host> host_;
+    std::shared_ptr<Router> router_{nullptr};
```
ThomasRuijters: Funny why is  this a shared pointer? What makes it shared? How come its lifetime is determined by more then one object? Are we going to have more then one multiplayer service?

src/network/client.cpp
ThomasRuijters: abra

src/network/client.cpp
ThomasRuijters: kadabra

src/network/host.cpp
ThomasRuijters: Wipety woosh

src/network/multiplayer_service.cpp
```diff
@@ -1,41 +1,160 @@
 #include <engine/network/multiplayer_service.h>
 
-void MultiplayerService::register_handler(const MessageType type, const std::function<void()>& handler)
+#include <stdexcept>
+
+MultiplayerService::MultiplayerService()
 {
+    if (enet_initialize() != 0) {
```
ThomasRuijters: What is this supposed to be? !=0 ? Isn't it just false? What is this supposed to mean? Isn't there an enum for this at least??? Please tell me there is...

src/main.cpp
```diff
@@ -11,7 +12,7 @@ namespace {
     #endif
 }
 
-int main() {
+int main(int argc, char* argv[]) {
```
ThomasRuijters: annnddd why did we add this?

ThomasRuijters (changes requested):
Looks really solid. There are still some remarks that I would like to give. Please make a custom CapycoreMultiplayerException class. You have so many places where you can throw exceptions that I believe it wise to make a general custom one, just for the sake of having the type be a bit more descriptive. Other than that it looks good. Some its a kind of magic vibes here and there but you can bump those out.

src/main.cpp
SjulGr: Leftovers from test code I forgot to remove, good spot!

src/network/multiplayer_service.cpp
SjulGr: It's code according to the [ENet tutorial](http://enet.bespin.org/Tutorial.html#Initialization). It doesn't state in the tutorial, but I'll take a look into it.

include/engine/network/connection_state.h
SjulGr: @BotSquirrel Other enum classes like `DefaultMessage` and `CustomMessage` are required for them to be converted into said type (e.g. the first of `DefaultMessage` will be converted to `0U` when interpreted as a string). `ConnectionState` doesn't need to be since it's a local value that won't be used as anything other than the `ConnectionState` enum.

In short, I could add a type in the name of consistency, but it would only add unused functionality.

src/network/multiplayer_service.cpp
SjulGr: `is_multiplayer` might be a nice addition in the future if it works well with other features we might want to add, but for now I'll remove the else statement.

include/engine/network/network_message.h
SjulGr: I'll take a look into this in the next task of the user story in which I'm going to implement custom message types which. It would be more appropriate to do this in that task I think.

BotSquirrel (approved):

ThomasRuijters (approved):

PR #114, opened 2025-11-22
feat (UI): renderering UI text and button

BotSquirrel:
This adds the rendering of UI text and buttons based on the renderer refactor and the original diagram from Brightspace. See the image below to what can be achieved now.

A couple notes:
* I tested this a lot with Tracy to check if fonts do not leak/re-load into memory, and that is not the case due to the expanding of assetserice.
* Buttons are not clickable yet as I/someone needs to add the Input System in the backend so I can track your mouse. It does register actions but does not fire them
* Tracy has a bug, which is not really a bug, that is also tackled here.... :|

image


BotSquirrel:
So I merged this back as its to a feature branche and it will be open for review anyhow. So i'm just switching its place :)

PR #118, opened 2025-11-24
feat (animator): add animator component

### LINKED ISSUE #117: Animation
Adds the ability to animate game objects using sprite sheets. Nothing more to say then that.

## This feature should result to:
Once this feature is done, the user should have the ability to add an animation component to a GameObject, pick a sprite sheet that is to be played, and control parts of the animation like starting, stopping, resetting, etc. The animation component is more of an animator, a controller of sorts.

ThomasRuijters:
1. Add the animator skeleton
2. Implement the animator skeleton
3. Remove the stop and start functions, there is a reset preset anyway. After some thinking I removed them, even though I added them in my own original design. They feel like bloat and can be removed.

### DIFF
--- a/src/public/animator.cpp
+++ b/src/public/animator.cpp
@@ -83,7 +83,11 @@ int Animator::calculate_next_frame_index(const int intervals_advanced) const {
 }
 
 void Animator::update_sprite_texture(const int new_frame_index) {
-    const auto& parent = this->parent().value().get();
+    const auto& parent_opt = this->parent();
+    if (!parent_opt.has_value()) {
+        throw std::runtime_error("Animator: No parent GameObject found.");
+    }
+    const GameObject& parent = parent_opt.value().get();
     const auto sprite = parent.get_component<Sprite>();
 
     if (!sprite.has_value()) {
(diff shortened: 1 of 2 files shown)


BotSquirrel (approved):
Nice, calculating time like this is better than keeping a timer

src/public/animator.cpp
```diff
@@ -0,0 +1,133 @@
...
+
+/**
+ * So the C++ modulo works crappy with negative numbers.
+ * Imagine current_frame_index = 2, frame_count = 5, frames_advanced = -3
+ * (2 + -3) = -1, -1 % 5 = -1 (what we do not want, invalid index)
+ * Add the frame count to make it positive: -1 + 5 = 4
+ * 4 % 5 = 4 (what we want)
+ * ((2 + -3) % 5 + 5) % 5 = (-1 + 5) % 5 = 4 % 5 = 4 (correct)
+ * This allows us to keep the code 'simple' ¯\_(ツ)_/¯.
+ */
```
SjulGr, line 53: I find this comment difficult to read. Since it's a single comment I'm talking about, I won't sign it as blocking and it's up to you if you want to change it, but it could use some better readibility.

SjulGr (approved):

PR #124, opened 2025-11-26
feat (go): dont destroy on load

### LINKED ISSUE #110: Keep marked game objects on scene switch
**Description:**
We need a mechanism that allows certain game objects to persist between scene loads, similar to Unity's `DontDestroyOnLoad`. When a scene switches, objects marked with this flag should not be destroyed but should persist in the new scene.

**Goal:**
- Implement functionality to mark game objects as "persistent" across scene transitions.
- Objects marked as `ddol` (DontDestroyOnLoad) should not be destroyed when loading a new scene.

**Acceptance Criteria:**
- **DDOL Flag:** Each game object should have a flag (e.g., `ddol`) that can be set programmatically or via the editor.
- **Scene Load Handling:** On scene load, game objects with the `ddol` flag should not be destroyed or unloaded.
- **Persistent Objects:** These game objects should remain active and should continue to function normally across scenes.
- **Edge Cases:**
  - Ensure that the ddol objects don't get duplicated when switching scenes.
  - Ensure that ddol objects are not destroyed if they are part of a new scene’s hierarchy.

**Potential Implementation Details:**
- During scene transition, mark game objects with the `ddol` flag to be excluded from destruction during the scene unload.
- Attach a system to check for `ddol` marked objects when a new scene is loaded and re-parent them to the new scene.

LarsDevans:


ThomasRuijters (changes requested):
No. Why do we want this? Me was not properly informed of this. Perhaps for performance sake but I do want this to be checked by the rest of the team.

BotSquirrel (approved):
I approve it, but I have to guess for now that your tests work as the pipeline fails on linter

ThomasRuijters (approved):

LarsDevans:
The linter is probably going to fail, therefore proof:
image

(The linter issues will be addressed with CI, thus skipping here)

PR #125, opened 2025-11-26
build: convert .clang-tidy to .clang-format with pre-commit hook

LarsDevans:
Including... all files... linted... Have mercy...


BotSquirrel (approved):
I don't see any logic changes (except CI) so I approve. As i've said I'm fine as long as everyone agrees, but i do see that this formats and not neccessarily corrects it. Things like copy value checks won't be taken in account so we might need to pay some more attention to guidelines in our reviews

ThomasRuijters (approved):

SjulGr (approved):

PR #126, opened 2025-11-26
feat (systemservice): add a new service that can interact with our backend (SDL)

BotSquirrel:
I seperated the main event loop from SDL in the `SystemService`. The reason why I called it a `System` service is because I intend to place some more methods there that belong to SDL but not to rendering specific. Things like framerate or the tick counter. The second one I already moved over.

The main loop would look something like this. *note: I didn't add this into scene as this would conflict a lot with the GUI branche, there it will be added.*

```cpp
void run() {
    auto& engine = Engine::instance();
    engine.initialize();

    bool running = true;
    float accumulator = 0.0f;
    float time_scale = 1.0f;
    float fixed_step = 1.0f / 60.0f;

    auto& system_service = Engine::instance().services->get_service<SystemService>().get();
    auto& input_manager = Engine::instance().services->get_service<InputManager>().get();

    int idx = system_service.add_listener(EVENT_QUIT, [&](void* /*event*/) { // register per scene
        running = false;
    });

    system_service.init_frame_timer();

    while (running) {
        system_service.update_frame_time(time_scale);
        float frame_dt = system_service.delta_time();
        accumulator += frame_dt;

        run_without_tracy([&]() {
            input_manager.update();
            system_service.update();
        });

        while (accumulator >= fixed_step) {
            // this creates a fixed update step if needed in the future for physics or similar
            accumulator -= fixed_step;
        }

        run_without_tracy([&]() {
            if (input_manager.provider().is_mouse_pressed(MouseButton::left)) {
                std::cout << "Mouse Left Button Held\n";
            }

            if (input_manager.provider().is_key_pressed(KeyCode::a)) {
                std::cout << "Key A Pressed\n";
            }

            if (input_manager.provider().is_key_pressed(KeyCode::m)) {
                std::cout << "Key m Released\n";
            }
        });
    }

    system_service.remove_listener(EVENT_QUIT, idx); // de-register on scene switch to avoid dupes.


    engine.quit();
}
```


ThomasRuijters (approved):

LarsDevans (approved):
Good solution

PR #128, opened 2025-11-26
feat (networking): add custom network message support

### LINKED ISSUE #54: Network expandability implementation
### Summary
Fill the skeleton for the `multiplayer service` with a working implementation to allow a developer to add their own message types and routes.

Goals
- Be able to register a new networking message type
- Be able to register a new networking router

### Deliverables
- Functionality to expand the `multiplayer service` outside of the engine

SjulGr:
Add the possibility to create your own set of network message types.

Closes #54

### DIFF
--- a/include/engine/network/network_message.h
+++ b/include/engine/network/network_message.h
@@ -4,6 +4,13 @@
 #include <vector>
 #include <cstring>
 
+/**
+ * @enum DefaultMessageTypes
+ * @brief Built-in protocol message types used for core connection events.
+ *
+ * These message types represent the minimal set of messages required
+ * for establishing and maintaining multiplayer sessions.
+ */
 enum class DefaultMessageTypes : uint16_t {
     CONNECT,
     DISCONNECT,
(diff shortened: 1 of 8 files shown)


include/engine/network/network_message.h
```diff
@@ -15,7 +14,52 @@ enum class DefaultMessageTypes : uint16_t {
 
 enum class CustomMessageTypes : uint16_t;
 
-using MessageType = std::variant<DefaultMessageTypes, CustomMessageTypes>;
+struct MessageType
```
BotSquirrel, line 38: Maybe some doc comments, or a form of documentation somewhere on the purpose/usage?

src/network/host.cpp
```diff
@@ -81,13 +81,16 @@ void Host::poll() noexcept
...
+
             msg.header.size =
-                event.packet->dataLength - sizeof(MessageType);
+                event.packet->dataLength - sizeof(raw_type);
 
             msg.payload.assign(
-                event.packet->data + sizeof(MessageType),
+                event.packet->data + sizeof(raw_type),
                 event.packet->data + event.packet->dataLength
             );
```
BotSquirrel, line 95: I do see some snippets like this one exist in multiple places. It would be nice to maybe create a util class for enet which has these? Might be overkill if this is the only function but perhaps something to think about if more duplicate enet snippets occur?

BotSquirrel (approved):

src/network/host.cpp
SjulGr: I'll keep this in mind for the future. I first want to make the networking module fully functional. When we have time left over, this could be implemented.

ThomasRuijters (approved):

PR #129, opened 2025-11-28
feat (UI): add interactable buttons and fix update loop

BotSquirrel:
This adds the interaction behind the buttons with an additional input field. Both of them now inherit from an interactable which makes it easier to add/tweak things later. This does not have any tests as I can't really test components and SDL logic...so here here is the result.

https://github.com/user-attachments/assets/64fafacb-3fa9-40a7-bef5-168cfd3a1c0e


src/core/rendering/strategies/sdl/sdl_text_strategy.cpp
ThomasRuijters, line 1: This one function is a 100+ lines. I would like to see this be split up into separate functions in the future. I won't block your PR for it but keep it in mind.

ThomasRuijters (approved):
It gets my sign of approval. One thing I want to generally mention is that I do not agree with your use of structs. This is not Rust and it isn't C, its C++. State lives with the behavior that affects it vice versa. OOP 101. But for now its fine I suppose, its a matter of taste.

LarsDevans (approved):
Great demonstration video

PR #131, opened 2025-11-30
feat (renderer): add basic layered rendering

### LINKED ISSUE #123: Layered rendering
Add sprite layering to rendering logic

ThomasRuijters:
Adds basic layering to the renderer. Do mind that the highest layer is rendered first. 0 is the background, infinity is the closest to the screen. A simple re-order can flip this if needed


LarsDevans:
This is not preferred (and required by our school) syntax for working with the sprite layers. Take a look at the provided UML (on Brightspace) and the proof of concept here: https://github.com/SPC-F/sprite-renderer-poc

ThomasRuijters:
> This is not preferred (and required by our school) syntax for working with the sprite layers. Take a look at the provided UML and the proof of concept here: https://github.com/SPC-F/sprite-renderer-poc

Sadly enough the diagram school provided is not enough. The abstraction of GUI objects was not taken into account. When you look at layering, you can not have layering on sprites and on GUI objects separately. Been there, done that, tried that, didn't work. You get into context conflicts => situations where the code will technically speaking always function, restricting the user is practically impossible, and the end result is 'invalid' state. You can't fact check if you do ordering in two places in a different way, that may or may not result in conflicts on the general thing that they all end up existing on. Basically what happens when two domains with a different perspective on something work with the same thing. It clashes. So setting the rendering layer on the renderables was the only solid way to guarantee that there are no domain clashes in the strategies.

src/public/scene.cpp
```diff
@@ -70,10 +72,16 @@ void Scene::game_loop() { // NOLINT [readability-make-member-function-const]
...
         auto game_objects = this->game_objects();
         for (auto game_object : game_objects) {
             for (auto component : game_object.get().get_components<Component>()) {
                 component.get().update(frame_dt);
+
+                if (Renderable * const renderable = dynamic_cast<Renderable*>(&component.get()); component.get().active()) {
+                    layered_renderables[renderable->rendering_layer()].push_back(*renderable);
+                }
             }
         }
```
BotSquirrel: This logic was moved down below in the `run_without_tracy` to avoid unneccessary allocations. Now we run the update twice.

BotSquirrel (commented):
Perhaps we should discuss the location of the `layering` value as a group? Our design can be different if we have a good reason and update the stakeholders. Personally I think both work. I do see another version where it being on a sprite would also work.

src/public/scene.cpp
ThomasRuijters: I agree. But it just depends on how you want to do it. I have tried 4 versions of this thing. Two before we had GUI objects, two with them. And it all came down to clashing. So when you set layering on sprites themselves, you give responsibility to the sprite strategy for layering them. Same as for GUI objects. The problem becomes that two strategies both try doing layering their own way. This is bad for multiple reasons. You can get clashes in definitions, clashes on the texture ordering with the renderer from SDL, a lot of shit can and will go wrong. This is why I decided to include layer as a property of the renderable itself. The user is just responsible for deciding where they want to layer something. This way saves us a lot of back pain. Its a shortcut, but one we can take due to the minimal design of the engine. Its a give or take situation. If you want to deal with re-ordering strategies (which would be the consequence if you do not use layering on the renderable) you could. But that would be a whole other beast because you would have to make a plug and play ordering pipeline as well, or risk inappropriate intimacy between strategies and the renderer. Trust me you do not want this design it makes it even more complex for reasons we just can agree to at the moment. Yes it can be done with layering via sprites and GUI objects separately => no you shouldn't do that. Bad idea.

For the update => A complete oversight on my part. Will look at it.

BotSquirrel:
@ThomasRuijters I checked the current diagram and see some tweaks that might make it easier/better:
`GameObject` has a layer, we could turn that into an `int` and use that for the buckets instead of renderable. Then sprites/rendering can have an order which dictates the order in that bucket itself. That way i can have something in LAYER and say that it should be rendered first in that bucket.

We can technically say that UIObjects will always be in the UI layer (perhaps an enum would also be nice for layers 🤔), so they don't neccessarily need a selection as why would i want that outside of UI? If you're wondering how you can fix that, in `scene` I handle `UIObject`'s differently as they require something special due to inheritence. You could add the push to bucket there. Alternatively as they are also gameobjects we can set their layer default to UI in constructors of those objects.

Both can work but I would like to utilize the layer in gameobject to indicate which bukket, and your current int usage as the order in the bucket. If that proves too much trouble then we can leave it as is for now and just move it to gameobject. 

The reason I want it on `GameObject` is the main diagram. Originally sprite has both but that leaves no purpose for the `layer` field on gameobject, and if we faithfully follow Unity, then its located on gameobject.

Quick side note; if you allocate something in memory in the main game loop in `Scene` place that in `run_without_tracy` so we don't log anything that gets allocated each frame.

ThomasRuijters:
> @ThomasRuijters I checked the current diagram and see some tweaks that might make it easier/better: `GameObject` has a layer, we could turn that into an `int` and use that for the buckets instead of renderable. Then sprites/rendering can have an order which dictates the order in that bucket itself. That way i can have something in LAYER and say that it should be rendered first in that bucket.
> 
> We can technically say that UIObjects will always be in the UI layer (perhaps an enum would also be nice for layers 🤔), so they don't neccessarily need a selection as why would i want that outside of UI? If you're wondering how you can fix that, in `scene` I handle `UIObject`'s differently as they require something special due to inheritence. You could add the push to bucket there. Alternatively as they are also gameobjects we can set their layer default to UI in constructors of those objects.
> 
> Both can work but I would like to utilize the layer in gameobject to indicate which bukket, and your current int usage as the order in the bucket. If that proves too much trouble then we can leave it as is for now and just move it to gameobject.
> 
> The reason I want it on `GameObject` is the main diagram. Originally sprite has both but that leaves no purpose for the `layer` field on gameobject, and if we faithfully follow Unity, then its located on gameobject.
> 
> Quick side note; if you allocate something in memory in the main game loop in `Scene` place that in `run_without_tracy` so we don't log anything that gets allocated each frame.

I agree to a certain extend. We should not make stuff more complex because we technically speaking have the code ready. Sprites have layers, game objects have layers. If we only use the layer of the game object, sprites do not need a layer. Simple as that. I am not adding complexity because we have an extra 3 lines of code in the app. If we do not use it, we just remove it. Less complexity is better. If it where up to me, I would skip the layering in renderables (thus also in sprites) and just use the one from the game object instead. A game object should not have more than one sprite anyway.

include/engine/core/rendering/renderable.h
```diff
@@ -13,14 +13,18 @@
...
- protected:
+class Renderable : public Component { // NOLINT
+protected:
   std::unique_ptr<IRenderingStrategy> render_strategy_;
-
- public:
+  int ordering_layer_ = 0;
+public:
   Renderable() = default;
+  explicit Renderable(int layer);
```
LarsDevans, line 22: Perhaps we should assign a default layer so the parameter could be omitted

LarsDevans (approved):

include/engine/core/rendering/renderable.h
```diff
@@ -13,14 +13,18 @@
...
- public:
-  Renderable() = default;
+  int ordering_layer_ = 0;
+public:
+  explicit Renderable();
+  explicit Renderable(int layer);
   ~Renderable() override = default;
 
+  Renderable& ordering_layer(int layer);
+  [[nodiscard]] int ordering_layer() const;
```
BotSquirrel: `ordering_layer` or `order_in_layer` as we depict our main orderin through the `layer` in gameobject?

src/core/engine.cpp
```diff
@@ -13,10 +13,13 @@
 #include <memory>
 
 Engine::Engine() : services(std::make_unique<ServiceContainer>()) {
+  services->register_service<SceneService>();
   services->register_service<RenderingService>();
   services->register_service<SystemService>();
   services->register_service<AudioService>();
   services->register_service<AssetService>();
+  services->register_service<InputManager>();
+  services->register_service<PhysicsService>();
```
BotSquirrel: Take a look at the FPS merge, the core 5 are registered here and input/physics are delayed in `initialize`

BotSquirrel (approved):
1 minor fix that is optional, but the engine one i'd like to see pass

PR #143, opened 2025-12-05
@feat/animation swapping

ThomasRuijters:


src/main.cpp
```diff
@@ -11,8 +12,48 @@ inline void tracy_shutdown() {}
...
 }  // namespace
 
+#include <engine/core/engine.h>
+#include <engine/public/scene_service.h>
+#include <engine/public/gameObject.h>
+#include <engine/public/components/sprite.h>
+#include <engine/core/rendering/assetService.h>
+#include "engine/public/components/animator.h"
+
+void yes() {
```
BotSquirrel: Isn't it better to keep main clean?

LarsDevans (approved):

BotSquirrel (approved):

src/main.cpp
ThomasRuijters: Al afgedekt :D

PR #144, opened 2025-12-05
fix (body2d): update outdated sizing calculation in members

BotSquirrel:
A small fix for the physics calculation in bod2D size members. Originally we had no conversion from pixel to box2D meters. I added this in a later PR. Now we must convert units before setting/getting metric values. This PR fixes that we set old low values.

I tested this myself using a behavior, it works as expected but shortens towards the top and not bottom so you'd need to counter it with an offset.

```cpp
auto box_collider_opt = get_component<BoxCollider2D>();

if (box_collider_opt.has_value()) {
  auto& box_collider = box_collider_opt->get();
  if (provider.is_key_held(KeyCode::c)) {
    box_collider.height(8.0f);  // crouch height
  } 
  else {
    box_collider.height(16.0f);  // normal height
  }
}
```

### DIFF
--- a/include/engine/physics/world/body/body_2d.h
+++ b/include/engine/physics/world/body/body_2d.h
@@ -3,6 +3,7 @@
 #include <box2d/box2d.h>
 #include <engine/physics/world/body/body_type_2d.h>
 #include <engine/physics/world/body/shape_type_2d.h>
+#include <engine/public/util/point.h>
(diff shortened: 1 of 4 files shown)


ThomasRuijters (approved):

ThomasRuijters:
Could you leave a short describtion next time?

LarsDevans (approved):

PR #145, opened 2025-12-06
refactor(go): allow scene manipulation from go

LarsDevans:
While working on the game I was required to do some manipulative operations on the scene. However, this is disallowed by the current setup

### DIFF
--- a/include/engine/public/gameObject.h
+++ b/include/engine/public/gameObject.h
@@ -75,7 +75,7 @@ class GameObject {
   [[nodiscard]] Transform& transform();
   [[nodiscard]] const Transform& transform() const;
 
-  [[nodiscard]] const Scene& scene() const noexcept;
+  [[nodiscard]] Scene& scene() const noexcept;
   void scene(Scene& scene) noexcept;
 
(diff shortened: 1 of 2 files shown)


ThomasRuijters:
What exactly were you doing? If you are required to change a scene you usually use the SceneService to get the current scene. You should not use GameObject instances for that unless required. Its not about an alternative way to do things, being consistent is more important here

BotSquirrel (commented):
I don't think its bad as it can be usefull for example time scaling which I imagine this might be for? My opinion on @ThomasRuijters comment above, I think that the scene service is the main manager responsible for scene creation and which is loaded. Scene manipulation is fine to a certain extend. It should not interfere with the loop or ordering. Altering time or adding an object is fine. 

Its also prone to lesser mistakes to just look at the `scene` member than ask the scene service for a specifc name which could have been changed. 

Before I approve it I think i'd be nice to have some form of documentation which dictates what a gameobject is allowed to do with it. If the user ignores that, then fine, his mistake.

SjulGr (approved):

BotSquirrel (approved):

Issue #137, opened 2025-12-05
AI


BotSquirrel:
### Summary
Create the AI components needed for graph generation, path generation and path following. This Task contains the creation of skeletons as the final implementation.

### Goals
- Be able to generate a graph from a parent gameobjects childeren
- Be able to generate the shortest path using a* between two points or a target
- Be able to move an AI agent towards a point

### Deliverables
The following components which can be placed on specific game objects:
* `Navigation graph` component
* `Navigation node` component
* `Pathfinding` component
* `AI controller` component

PR #148, opened 2025-12-08
feat (ai): add the implementation of the navigation graph generation

BotSquirrel:
See #137 for more information. TLDR; the implementation of the navigation graph generation. In order to make this work I also had to tweak some on `GameObject` (scene) and `Transform`. The position thing on `Transform` is going to get its own PR but as I said, I needed it here.

The component generates a graph based on the children of a parent, I also added a debug draw to see the path which can be seen. Down below is an example maze I created, which it accurately produced a graph for.

It does require a little bit of config as to set the grid size and strides it can take. I added the option for weights but don't do anything with it currently.

image

### DIFF
--- a/include/engine/core/rendering/renderable.h
+++ b/include/engine/core/rendering/renderable.h
@@ -1,9 +1,9 @@
 #pragma once
 
-#include <memory>
-
 #include <engine/core/rendering/strategies/irendering_strategy.h>
 #include <engine/public/component.h>
+
+#include <memory>
 
 /**
  * @brief Base class for renderable components in the engine.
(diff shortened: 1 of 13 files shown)


include/engine/core/rendering/renderable.h
```diff
@@ -27,4 +29,7 @@ class Renderable : public Component { // NOLINT
 
   void set_render_strategy(Component& component);
   [[nodiscard]] virtual IRenderingStrategy& render_strategy() const;
+
+  Renderable& draw(bool enabled) noexcept;
```
ThomasRuijters: Instead of a bool enabled write two setter functions DisableDraw and EnableDraw. Thats clearer than boolean passing imo.

src/core/rendering/renderable.cpp
```diff
@@ -25,4 +27,11 @@ void Renderable::set_render_strategy(Component& component) {
...
-}
\ No newline at end of file
+}
+
+Renderable& Renderable::draw(bool enabled) noexcept {
+  should_draw = enabled;
+  return *this;
+}
+
+bool Renderable::draw() const noexcept { return should_draw; }
```
ThomasRuijters: Perhaps a rename to ShouldDraw or IsDrawEnabled is better here.

ThomasRuijters (approved):

SjulGr (approved):

PR #152, opened 2025-12-10
feat (ai): pathfinding component

BotSquirrel:
See #137 for more information. TDLR; the pathfinding component which looks at the corresponding graph for nodes. It supports weights and slightly increases them for jumps and drops. It would be cool to expand it more but this suffices for now.

Also if someone comments that the main a* method is big, well yeah it is. Computing a path is no small feat and multiple members are constantly read or updated. I tried to keep it as small as i could and would prefer not to split it.

image


src/core/rendering/strategies/sdl/sdl_strategy_factory.cpp
```diff
@@ -43,5 +45,9 @@ std::unique_ptr<IRenderingStrategy> SdlStrategyFactory::create_strategy(
         *renderer_.sdl_renderer_);
   }
 
+  if (auto* pathfinding = dynamic_cast<Pathfinding*>(&component)) {
```
ThomasRuijters: const pointer perhaps?

src/public/components/ai/navigation/navigation_node.cpp
```diff
@@ -6,7 +6,7 @@ NavigationNode::NavigationNode() {
       throw std::runtime_error("Collider2D has no parent GameObject.");
     }
 
-    this->disable_draw();
+    // this->disable_draw();
```
ThomasRuijters: Not needed?

src/public/components/ai/navigation/pathfinding.cpp
ThomasRuijters, line 1: I won't block the PR for this, but perhaps it would be wise to split the functions up when they pass 100+ lines. Sometimes you can not name things or shove them inside another box so I get it if that's the case here.

ThomasRuijters (approved):

src/public/components/ai/navigation/pathfinding.cpp
BotSquirrel: See my note about this in the description. I agree with it but this is a case where its very difficult to move this apart without rewiring/doing things differently

SjulGr (approved):

PR #153, opened 2025-12-11
feat(networking): gamestate synchronization

### LINKED ISSUE #56: Networking gameobject updates
### Summary
Fill the skeleton for the multiplayer service with a working implementation to allow the engine to gather all networking objects data, serialize it and broadcast it to the clients to sync all clients with the current networking data.

### Goals
- Be able to send the entire game state from the host to all clients

### Deliverables
Functionality to keep the entire game in sync across multiple clients.

SjulGr:
Adds automatic synchronization to game objects via delta snapshots that includes 'dirty' game objects. Also adds an initial snapshot that include all game objects with a network identity.

Note about the video: The top square is the host. In the test program, the host doesn't broadcast when they move, but they do update their position. The top square demonstrates how frequent a game object is synced on other clients.

https://github.com/user-attachments/assets/bc938c9e-6954-40b0-b9de-aa2a93f2c116

Closes #56


include/engine/public/component.h
```diff
@@ -2,8 +2,14 @@
 #include <functional>
 #include <memory>
 #include <optional>
+#include <string>
+#include <typeinfo>
 #include <vector>
 
+#include <cstddef>
+
+namespace std { class type_info; }
```
BotSquirrel: There is an import for type_info? `#include <typeinfo>`

include/engine/public/components/network_identity.h
```diff
@@ -0,0 +1,41 @@
...
+    void set_owner_uuid(const std::string& owner) noexcept;
+
+    void mark_dirty() noexcept;
+    void clear_dirty() noexcept;
+    [[nodiscard]] bool is_dirty() const noexcept;
+
+    // Component interface
+    void update(float /*dt*/) override {}
+    void on_attach() override {}
+    void on_detach() override {}
```
BotSquirrel: Technically it does not need this as I swapped it for hooks in previous PR's

include/engine/public/components/network_identity.h
```diff
@@ -0,0 +1,41 @@
...
+
+#include <string>
+
+/**
+ * @brief Component that gives a GameObject a network identity (UUID) and dirty flag.
+ *
+ * Attach this to any GameObject that should be replicated. Systems or user code
+ * should call `mark_dirty()` when the object's networked state changes.
+ */
+class NetworkIdentity : public Component {
```
BotSquirrel, line 13: Does this work for non-player entities aswell? If not then it might be good to specifiy that here

src/public/gameObject.cpp
```diff
@@ -121,6 +133,15 @@ std::vector<std::reference_wrapper<GameObject>>& GameObject::children() {
   return children_;
 }
 
+std::vector<std::reference_wrapper<Component>> GameObject::get_components_all() const {
```
BotSquirrel, line 136: You could call `get_components<Component>` to get the exact same result. I don't mind an extra function but there was an existing variant.

src/public/gameObject.cpp
```diff
@@ -135,11 +156,108 @@ GameObject& GameObject::remove_child(GameObject& child) {
   return *this;
 }
 
-// NOLINTBEGIN
-void GameObject::serialize() const {
-  throw std::runtime_error("Not implemented");
+void GameObject::serialize(std::vector<uint8_t>& out) const {
```
BotSquirrel, line 159: Some documentation on how to do serialize/deserialize in components might not be a bad idea. A simple example of how to set/read a value perhaps?

src/public/prefab_registry.cpp
```diff
@@ -0,0 +1,44 @@
+#include <engine/public/prefab_registry.h>
+#include <engine/public/scene.h>
+
+#include <stdexcept>
+
+PrefabRegistry& PrefabRegistry::instance() {
+  static PrefabRegistry singleton;
+  return singleton;
+}
```
BotSquirrel: I am not the biggest fan of this being another singleton. Preferably I'd see it as a service or extension on the asset service, but its not breaking so I won't block.

BotSquirrel (changes requested):

include/engine/network/multiplayer_service.h
```diff
@@ -65,6 +77,12 @@ class MultiplayerService : public IEngineService {
      * @brief Sends a message. Host broadcasts; client sends to server.
      */
     void send(const Message& message);
+    /**
+     * @brief Sends a message to a specific connected client (host mode only).
+     * @param uuid Target client's UUID.
+     * @param message Message to send.
+     */
+    void send_to_uuid(const std::string& uuid, const Message& message);
```
ThomasRuijters: You do not send to a UUID, you send to a connected client or host so probably a network participant which in your case is a peer. Make it explicit what you are doing here I can not send data to a UUID

include/engine/public/component.h
ThomasRuijters: #include <typeinfo> and a forward declare of class type_info?

include/engine/public/component.h
```diff
@@ -30,8 +36,13 @@ class Component {
...
 
-  virtual void on_serialize() = 0;
-  virtual void on_deserialize() = 0;
+  virtual void on_serialize(std::vector<uint8_t>& /*out*/) const;
+  virtual void on_deserialize(const std::vector<uint8_t>& /*data*/,
+                                       size_t& /*offset*/);
+
+  // A stable-ish textual identifier for the component type. By default this
+  // returns the RTTI name; components may override to provide nicer names.
+  virtual std::string type_name() const;
```
ThomasRuijters: stable-ish? What is the ish part here?

src/network/host.cpp
```diff
@@ -195,6 +207,31 @@ void Host::send_to_peer(const Message& message, ENetPeer* peer) noexcept
     enet_peer_send(peer, 0, packet);
 }
```
ThomasRuijters, line 208: perhaps you want to split this function up? Its a 150+ lines

src/network/host.cpp
```diff
@@ -195,6 +207,31 @@ void Host::send_to_peer(const Message& message, ENetPeer* peer) noexcept
...
 
+void Host::send_to_uuid(const std::string& uuid, const Message& message) noexcept
```
ThomasRuijters: Again I do not send stuff to UUIDs, I send information to peers. Does this thing also make sure you can not send information to yourself?

src/network/multiplayer_service.cpp
ThomasRuijters, line 1: You use implicit conversions a lot. If(_host), if(client), that is cool and all but it does not tell me anything. When I read that I can not tell what null or empty means there. Is it connected? Disconnected? Is it about signals being send that have to be mapped accordingly? What am I looking at here?

src/network/snapshot.cpp
```diff
@@ -0,0 +1,253 @@
...
+    uint8_t* write_ptr = payload.data() + old;
+    std::memcpy(write_ptr, uuidbuf, UUID_LEN);
+    write_ptr += UUID_LEN;
+    if (!obj_payload.empty()) {
+        std::memcpy(write_ptr, obj_payload.data(), obj_payload.size());
+    }
+}
+
+// Helper: Create snapshot message with a filter predicate
+template<typename FilterFunc>
```
ThomasRuijters, line 92: Could you use concepts here to enforce a certain type? What is the expected signature of FilterFunc?

src/network/snapshot.cpp
```diff
@@ -0,0 +1,253 @@
...
+    // Write count at the beginning
+    std::memcpy(payload.data(), &count, sizeof(count));
+
+    msg.payload = std::move(payload);
+    msg.header.size = static_cast<uint32_t>(msg.payload.size());
+    return msg;
+}
+
+// Helper: Find or create GameObject by UUID
+static GameObject* find_or_create_object(Scene& scene, const std::string& uuid_str,
```
ThomasRuijters: Why are we returning pointers instead of unique pointers?

src/network/snapshot.cpp
```diff
@@ -0,0 +1,253 @@
...
+        return true;
+    };
+
+    // Skip object metadata
+    if (!skip_string()) return;           // prefab_type_id
+    if (!skip_string()) return;           // name
+    if (!skip_string()) return;           // tag
+    if (!skip_bytes(sizeof(uint8_t))) return;   // is_active
+    if (!skip_bytes(sizeof(int32_t))) return;   // layer
+    if (!skip_bytes(sizeof(float) * 7)) return; // transform (7 floats)
```
ThomasRuijters, line 216: skip_if_invalid?

src/public/gameObject.cpp
ThomasRuijters: The intention is clearer with the new version. I like his approach more it communicates with words instead of mechanics

src/public/gameObject.cpp
```diff
@@ -135,11 +156,108 @@ GameObject& GameObject::remove_child(GameObject& child) {
...
+  auto pos = transform_.position();
+  float rot = transform_.rotation();
+  auto sc = transform_.scale();
+  snapshot::write_bytes(out, &pos.x, sizeof(float));
+  snapshot::write_bytes(out, &pos.y, sizeof(float));
+  snapshot::write_bytes(out, &pos.z, sizeof(float));
+  snapshot::write_bytes(out, &rot, sizeof(float));
+  snapshot::write_bytes(out, &sc.x, sizeof(float));
+  snapshot::write_bytes(out, &sc.y, sizeof(float));
+  snapshot::write_bytes(out, &sc.z, sizeof(float));
```
ThomasRuijters, line 181: this is fine, I won't block for it, but consider using just sizeof(var name). Unless the type you want to write is different from the declared or actual type of the variable, just use sizeof(variable)

src/public/gameObject.cpp
```diff
@@ -135,11 +156,108 @@ GameObject& GameObject::remove_child(GameObject& child) {
...
+  }
+
+  uint16_t comp_count = static_cast<uint16_t>(comp_entries.size());
+  snapshot::write_bytes(out, &comp_count, sizeof(comp_count));
+
+  for (const auto& e : comp_entries) {
+    snapshot::write_string(out, e.first);
+    uint32_t plen = static_cast<uint32_t>(e.second.size());
+    snapshot::write_bytes(out, &plen, sizeof(plen));
+    if (plen > 0) {
```
ThomasRuijters: plen?

ThomasRuijters (changes requested):

include/engine/public/component.h
SjulGr: `type_name` used to make use of compile-time assigned names, which was stable when using the same compiler for every build. On second inspection I changed the implementation of this so each component implements their own type_name() due to the fact the we do make use of different compilers.

include/engine/public/components/network_identity.h
SjulGr: It does work for non-player entites. Everything that should be synced across the network should have the `NetworkIdentity` component. All you need to do is mark the object as dirty as soon as the state changes and it will be synced along with the player entities.

src/public/gameObject.cpp
SjulGr: The helper functions in `snapshot.h` provide comments on how to use said helpers. Using the serializer function in `gameObject.h`, you can see how a specific game objects gets serialized and deserialized. How do you recommend on improving the comments seeing as it already contains, in my opinion, clear comments?

src/network/host.cpp
SjulGr: I think you misread. Your comment does mark 150+ lines, but you're also marking four functions:
- `Host::poll()`
- `Host::broadcast()`
- `Host::disconnect()`
- `Host::send_to_peer()`

src/public/gameObject.cpp
SjulGr: While writing `sizeof(variable)` does make sense, but using explicit type names shows the intent better, ensures type safety and serialization stability. So using the variable does work, but could produce unexpected behavior.

src/public/gameObject.cpp
SjulGr: Payload length. I'll make the names more clear.

src/network/snapshot.cpp
SjulGr: I don't see how this would improve the code. We don't need to skip if invalid, we need to skip the size of said value. the skip functions return `false` if by skipping said value, we move out of the data size, so we're done with the object.

src/network/host.cpp
SjulGr: Correct, this method looks through all connected clients and sends it if client with given uuid is found. Otherwise, it does nothing.

src/network/multiplayer_service.cpp
SjulGr: It's basically as following:

`If` I am registered as `client`, execute `method`.
`Else if` I am registered as `host`, execute `method`.
`Else` throw exception that I'm not registered as either. Whether or not you're connected doesn't matter. That is handled within the implementation of `method`.

src/network/snapshot.cpp
SjulGr: The current code works best for the functionality. I added comments so you can see what is to be expected.

ThomasRuijters (approved):

BotSquirrel (approved):

PR #155, opened 2025-12-13
feat (ai): add the other components and final controller

BotSquirrel:
See #137 for more information. TDLR; This adds the final AI controller with all the other components. All except the AI Controller have already been reviewed in other PR's, so the focus lies on the controller.

I added 2 modes for now: Patrol and Chase. Both do exactly as they say. I also thought about adding a dynamic variant but that would take a lot more time and this suffices for now. The game developer also has the option to switch this.

Both modes work for kinetic (non-physics) objects and objects with a rigidbody, which you can see here below. I tried to do the rigidbody as best and simple as I could. Personally I think its looks great considering the time and effort but the occusional jitter can occur on higher speeds. Its best that we apply a knockback/kill something too avoid this.

If we actually want to move an entity then it might work, but I am not certain. We never talked about this, and if we do want it then I could always make a fix branche for it later that tweaks some of the physics.

For our game this should work fine, I also added a non-graph option so that a drone can fly straight at you :)

https://github.com/user-attachments/assets/b041674a-5b13-4d18-b70e-ed60472acac9


SjulGr (approved):

include/engine/public/components/ai/navigation/graph.h
```diff
@@ -0,0 +1,67 @@
...
+
+/**
+ * @brief Hash function for GraphPosition to be used in unordered containers.
+ * Combines the hash values of the x and y coordinates to produce a unique hash.
+ */
+struct GraphPositionHash {
+  size_t operator()(const GraphPosition& p) const noexcept {
+    return (std::hash<int>()(p.x) << 1) ^ std::hash<int>()(p.y);
+  }
+};
```
ThomasRuijters, line 45: Why a callable struct instead of just a function? And why not make it a member function on the GraphPosition instead if it stores no effects on the functor? This is not rust we do not do traits here.

include/engine/public/components/ai/navigation/graph.h
```diff
@@ -0,0 +1,67 @@
...
+/**
+ * @brief Record structure used in A* pathfinding algorithm.
+ *
+ * @param pos The graph position of the node.
+ * @param g The cost from the start node to this node.
+ * @param h The heuristic cost estimate from this node to the target node.
+ * @param parent The graph position of the parent node in the path.
+ * @param has_parent Flag indicating whether this node has a parent.
+ *
+ * @return f() The total estimated cost (f) from start to target through this
```
ThomasRuijters, line 56: call it cost then. F tells me nothing, cost does. We are not doing math here do not min-max keystrokes please. Names are meaning, be explicit we are writing language not scrambled symbolic.

include/engine/public/components/ai/navigation/navigation_graph.h
```diff
@@ -0,0 +1,136 @@
...
+   */
+  NavigationGraph& generate_graph();
+
+  /**
+   * @brief Clears all nodes from the navigation graph.
+   * @return Reference to the NavigationGraph instance for method chaining.
+   */
+  NavigationGraph& clear();
+
+  std::unordered_map<GraphPosition, std::reference_wrapper<NavigationNode>,
```
ThomasRuijters, line 45: Do we want to expose a read write version of this map? Or is it intended to be read-only?

include/engine/public/components/ai/navigation/pathfinding.h
```diff
@@ -0,0 +1,105 @@
...
+/**
+ * @brief Component responsible for AI pathfinding using a navigation graph.
+ *
+ * This component generates paths for AI-controlled GameObjects to navigate
+ * through the environment based on a NavigationGraph.
+ *
+ * NOTE: This component requires a NavigationGraph component to be present
+ * on the same GameObject to function correctly.
+ */
+class Pathfinding : public Renderable {
```
ThomasRuijters, line 18: The name could perhaps be a bit different. If it is a component that generates paths for AI controlled gameobjects to navigate is it then a Pathfinding or a PathGenerator? Naming matters here.

src/public/components/ai/navigation/pathfinding.cpp
```diff
@@ -0,0 +1,269 @@
...
+
+  auto heuristic = [&](const GraphPosition& a, const GraphPosition& b) {
+    return Point(a.x, a.y).distance_to(Point(b.x, b.y)).x;
+  };
+
+  NodeRecord start;
+  start.pos = origin;
+  start.g = 0.f;
+  start.h = heuristic(origin, target);
+  start.has_parent = false;
```
ThomasRuijters, line 142: Again I won't shoot you down for this, but please use words not letters if they carry meaning. g and h tell me nothing. Not min-maxing keystrokes speak English not math

ThomasRuijters (approved):
I won't shoot down your PR for the minor thingies I remarked, but do take a look at it if you find the time.

include/engine/public/components/ai/navigation/graph.h
BotSquirrel: `std::unorderd_map` does not allow that. It has to be its own individual thing and can't be combined with graph position. Struct is a perfectly valid option here, its also not a trait as that would mean that it would answer some question like `is_graph_position`. Its just a unique hashing object for easier and faster searching.

include/engine/public/components/ai/navigation/graph.h
BotSquirrel: Sure thats my bad and I'll fix it, however this was from earlier PR's and not the main focus of this one

include/engine/public/components/ai/navigation/navigation_graph.h
BotSquirrel: Intended read-only. Technically I left a small open window for others to interact with this using behaviors to create some variation but its main purpose is to be used by components and not overriden.

include/engine/public/components/ai/navigation/pathfinding.h
BotSquirrel: The component has an origin and a target. It generates a path based on that, nothing more and nothing less. In my head that is a pathfinder, hence the name pathfinding.

PR #170, opened 2025-12-21
feat: add `clone` method to `GameObject` and `Camera` classes to support object duplication

LarsDevans:


include/engine/public/camera.h
```diff
@@ -24,4 +24,9 @@ class Camera final : public GameObject {
 
   [[nodiscard]] int get_screen_width() const;
   [[nodiscard]] int get_screen_height() const;
+
+  [[nodiscard]] std::unique_ptr<GameObject> clone() const override {
+    return std::make_unique<Camera>(scene(), background_color(), zoom(),
+                                    is_main());
+  }
```
BotSquirrel, line 31: Definiton in `.h`?

src/public/gameObject.cpp
```diff
@@ -265,4 +265,12 @@ void GameObject::deserialize(const std::vector<uint8_t>& data, size_t& offset) {
...
-// NOLINTEND
\ No newline at end of file
+// NOLINTEND
+
+std::unique_ptr<GameObject> GameObject::clone() const {
+  // Cloning a complex object is.. well.. complex. Please avoid, since
+  // there's little time to implement that. The goal of the clone is to
+  // assist with the ItemDropper feature.
+  // Sincerely, Lars :)
+  return std::make_unique<GameObject>(scene());
```
BotSquirrel, line 275: does this actually clone, I feel like it just spawns an empty? Correction: I see what you do in the game ;). Perhaps make a note that it should be overriden so that is clear here.

BotSquirrel (changes requested):

PR #174, opened 2025-12-22
feat (multiplayer): send messages as udp instead of tcp

SjulGr:
Change from reliable (TCP) to unreliable (UDP) due to speed reasons


BotSquirrel (approved):
Odd how they didn't make a flag for this, as the 0 makes it unclear if you do not know the context

ThomasRuijters (approved):

PR #183, opened 2026-01-02
feat: fix scene load stack frame nesting

LarsDevans:
Questions are for @ThomasRuijters ;)

### DIFF
--- a/src/public/scene.cpp
+++ b/src/public/scene.cpp
@@ -14,6 +14,7 @@
 #include <engine/public/scene.h>
 #include <engine/public/ui/ui_object.h>
 #include <engine/util/memory.h>
+#include <engine/public/scene_service.h>
 
 #include <algorithm>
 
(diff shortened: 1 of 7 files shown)


SjulGr (approved):

BotSquirrel (changes requested):
I found 2 bugs 🐛, which have been resolved:
* Can switch from A to B but not from B to A due to it always being flagged as `marked_for_stopping`
* Cannot quit the scene_service due to it never stopping the current scene.

I will approve this, however I want to see a branche on the game with a tweaked CMake config that pulls this branche. That way main is intact on both the Engine and Game while we can test the game.

If that branche works as intended then I will approve both and can we merge in sequence.

src/public/scene.cpp
```diff
@@ -167,20 +175,18 @@ void Scene::run() {
...
       Engine::instance().services->get_service<SystemService>().get();
   stop_event_listener_id_ = system_service.add_listener(
-      EVENT_QUIT, [&](void* /*event*/) {  // register per scene
-        stop();
-      });
+  EVENT_QUIT, [&](void* /*event*/) {  // register per scene
+    stop();
+    Engine::instance().services
+      ->get_service<SceneService>().get()
+      .stop();
```
BotSquirrel, line 182: `SceneService.stop()` calls `scene.stop()` under the hood so this makes it double. Not necessarily an issue but not needed

BotSquirrel (approved):

PR #185, opened 2026-01-04
fix (gameobject): set transform parent on set parent

BotSquirrel:
This PR fixes the problem that gameobjects which get a parent don't always take their transform parent causing the wrong location display. An example was picking and dropping a weapon in quick succession, the speed of it could cause it not being set. Now its always enforced as we know a parent is present.


LarsDevans (approved):
Seems to work just fine!

ThomasRuijters (approved):

PR #188, opened 2026-01-08
feat(ai): add chase player

LarsDevans:
> [!IMPORTANT]
> Related issue: https://github.com/SPC-F/Capybrawla/pull/109

### DIFF
--- a/src/public/components/ai/ai_controller.cpp
+++ b/src/public/components/ai/ai_controller.cpp
@@ -1,6 +1,12 @@
+#include <engine/public/behavior.h>
 #include <engine/public/components/ai/ai_controller.h>
+#include <engine/public/gameObject.h>
+#include <engine/public/scene.h>
+#include <engine/public/transform.h>
+#include <engine/public/util/vector3.h>
 
 #include <algorithm>
+#include <iostream>
 #include <stdexcept>
 
 std::reference_wrapper<Pathfinding> AIController::get_pathfinding_component() {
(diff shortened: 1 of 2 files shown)


src/public/components/ai/ai_controller.cpp
```diff
@@ -40,6 +46,39 @@ AIController::AIController(GameObject& pathfinding_game_object)
...
+          }
+        }
+      }
+
+      if (closest_player) {
+        set_chase_target(*closest_player);
+        set_mode(AIControllerMode::CHASE);
+      }
+    }
+  }
```
BotSquirrel, line 80: This should be game logic, not engine logic. Besides, with this we hardcode the AIController to only chase players. The callbacks are fine and public. You could write a behavior that checks this logic and sets the mode/chase target. If you want it fancy you could also make a sensor circle around the AI, if the player is within range (the circle) then target.

src/public/components/ai/ai_controller.cpp
```diff
@@ -70,7 +109,14 @@ void AIController::chase(float dt) {
...
 
-    if (dist < arrival_threshold_) return;
+    if (dist < attack_distance_) {
+      if (attack_timer_ <= 0) {
+        for (auto& action : on_chase_threshold_reached_actions_) {
+          action(*this);
+        }
+        attack_timer_ = attack_cooldown_;
+      }
+    }
```
BotSquirrel, line 119: Same here. A callback on reach is fine, but specifying that this can attack should not be here. The AI controller should just be a component from the engine that handles internal pathfinding logic. Ideally you'd write your own controller in the game and use these in combination. So add two behaviors: `AIController` and `EnemyController` or something like it

The AIController moves and your own controller detects range and applies damage

src/public/components/ai/ai_controller.cpp
```diff
@@ -135,6 +181,20 @@ void AIController::try_traverse_graph(Transform& source, Transform& target,
           action(*this);
         }
       }
+    } else if (mode_ == AIControllerMode::CHASE) {
+      float dist_to_target = (source_center - target_center).length();
+      std::cout << "CHASE1 dist: " << dist_to_target
```
BotSquirrel, line 186: Left over `cout`'s

BotSquirrel (commented):
Mhm I don't really know what to say at this point in time. Personally I would like to see this changed, but I first want to know what others think.

LarsDevans:
No engine code is touched with the update in https://github.com/SPC-F/Capybrawla/pull/109

